Issue #16, opened 2019-11-04
Invulnerables (HotFix)


jordy25519:
As a node operator I want some network peers to be invulnerable to the reputation system so that my network is less volatile.

Tasks:
- [x] Make a branch from cennznet/plug-blockchain ~`9d72ac28731d2dbf558f21683d17bd4f799213ca`~ `stable` for the hotfix work

- [x] Change  `on_report_peer` behavior so that reserved node connections are not dropped
https://github.com/plugblockchain/plug-blockchain/blob/9d72ac28731d2dbf558f21683d17bd4f799213ca/core/peerset/src/lib.rs#L261

- [x] Write integration tests to mock CENNZnet Rimu network structure
https://github.com/plugblockchain/plug-blockchain/tree/develop/core/network/src/test

- [x] __A/C__ Hotfix deployed by SREs into CENNZnet Rimu network

mzxyz:
Start working on the first 2 tasks

hoani:
Investigated the reputation mechanism. Came to the conclusion that reputation is not broadcast, therefore a unit test is adequate.

cowboy-bebug:
Yeah I think it makes sense for each node to keep its own local copy of the ranking / reputation (because different nodes probably have different reputation for the same peer)

PR #19, opened 2019-11-05
#16 Invalnerables: task 1,2

mzxyz:
#16

### DIFF
--- a/core/cli/src/params.rs
+++ b/core/cli/src/params.rs
@@ -96,7 +96,7 @@ pub struct NetworkConfigurationParams {
 	#[structopt(long = "bootnodes", value_name = "URL")]
 	pub bootnodes: Vec<String>,
 
-	/// Specify a list of reserved node addresses
+	/// Specify a list of reserved node addresses. Reserved node is invulnerable to reputation changes.
 	#[structopt(long = "reserved-nodes", value_name = "URL")]
 	pub reserved_nodes: Vec<String>,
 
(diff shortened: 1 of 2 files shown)


cowboy-bebug (commented):
Nice and clean 👍

jordy25519:
This should do the trick 👍 

- Can you write a unit test? for `on_report_peer`
e.g. when the peer id is reserved to verify the behaviour

- Add a comment here about reserved nodes becoming invulnerable to reputation changes
This will print in the help message i.e `plug --help`

https://github.com/plugblockchain/plug-blockchain/blob/3d533cf343484ad56d18135e84338b51942a3646/core/cli/src/params.rs#L108

hoani:
It would be great to add a unit test similar to test_peerset_banned(). In the test we would want show that the node is never banned if it is reserved.
We can ignore the part of test_peerset_banned where we wait for the reputation to recover =)

CI seems to be failing due to a dependency issue - I wrote this guide https://centralitydev.atlassian.net/wiki/spaces/WHO/pages/761102339/Setting+Up+Local+CircleCi+for+Mac which lets us run circle ci on our own PCs without pushing to github (although it will still take ages)

cowboy-bebug:
I think we need to go ~further back~ forward in history. `cennznet/master` requires `srml-consensus` module which was removed in this commit dfe6779dcab59a6b52d61bbe53fdf414f3fc887d

bc276f3c9c36652e738ef1c7f705a5faefd36640 added srml/consensus
86a1dccecf90dd1f7bffb859b629bf295bbfa01b refactored and added some more stuff to srml/consensus <- so maybe use this one?

Edit:
@HaoZhaiX Could u rebase to this commit: ~2b43a43dd0dca9a4bb2e3b3253af0e44f71a56ba~ 86a1dccecf90dd1f7bffb859b629bf295bbfa01b

cowboy-bebug:
Closing as the issue is fixed in the legacy repo.

PR #27, opened 2019-11-14
Remove all references to flaming fir (#21)

### LINKED ISSUE #21: Remove flaming fir references
Plug node is inoperable with the flaming fir testnet so all references should be removed.

e.g. https://github.com/plugblockchain/plug-blockchain/blob/cfc1eaac64b637d2884da91c49d34e8940c8e993/node/cli/src/lib.rs#L65

aliXsed:
Closes https://github.com/plugblockchain/plug-blockchain/issues/21


node/cli/src/lib.rs
```diff
@@ -53,16 +53,13 @@ pub enum ChainSpec {
 	/// Whatever the current runtime is, with simple Alice/Bob auths.
 	LocalTestnet,
 	/// The Flaming Fir testnet.
```
jordy25519: Should remove this comment and keep the one below

jordy25519 (approved):
LGTM! assuming CI passes

Can change comment to say "Closes #21" and github automation will close the issue on merge

node/cli/src/lib.rs
aliXsed: Will do

cowboy-bebug (approved):
Yep looks good!

PR #30, opened 2019-11-20
reinstate reward method, removed in upstream merge

cowboy-bebug:
- Adding `reward()` method back, which was removed in the upstream merge.
- We need it for `cennzx-spot` module


cowboy-bebug:
creating a new PR

PR #35, opened 2019-12-06
Merge/upstream

jordy25519:
Latest round of upstream changes

The usual misc. bug fixes, reverts, refactors, renames etc.

Notably includes:
- The "big reorg" where `srml` was renamed to frame / palette. Don't quite understand the naming but we've still got `prml`
- Important fix to prevent block production stalling: https://github.com/paritytech/substrate/pull/4215 https://github.com/paritytech/substrate/pull/4202
- grats to @elim0322 whose changes are included in this too 🚀


aliXsed (approved):
Basically we can only review our own changes which are limited and LGTM.

jordy25519:
It's just these UI errors, the output is tied to the rustc version so they break every so often.
I'm not sure which rustc version they're supposed to work with is the problem.
Screen Shot 2019-12-09 at 1 05 23 PM

cowboy-bebug:
> It's just these UI errors, the output is tied to the rustc version so they break every so often.
> I'm not sure which rustc version they're supposed to work with is the problem.

Right let me ask on their matrix chat to confirm.

Update: asked here https://matrix.to/#/!HzySYSaIhtyWrwiwEV:matrix.org/$15759219801715469hcuZe:matrix.org?via=matrix.parity.io&via=matrix.org&via=web3.foundation

Otherwise we could `#[ignore]` some of these tests

PR #38, opened 2019-12-19
Keep origin of contract call stack

### LINKED ISSUE #34: Keep extrinsic sender through contract call stack
Provide the means to query the extrinsic sender address during nested contract execution.
So that the extrinsic origin is known throughout nested contract calls.

For reference, `caller()` gives the address of the calling account (extrinsic sender at the top level, changes to calling contract for nested calls)
https://github.com/plugblockchain/plug-blockchain/blob/b93e3585eb9bb9f1648a0d155e67832a8aa7f30c/srml/contracts/src/exec.rs#L139-L140

unblocks: #32

aliXsed:
Closes #34

### DIFF
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -142,6 +142,9 @@ pub trait Ext {
 	/// Returns a reference to the account id of the caller.
 	fn caller(&self) -> &AccountIdOf<Self::T>;
 
+	/// `origin` is the account which authored the original extrinsic leading to the current contract call(s). It does not change on nested calls as with `caller`.
+	fn origin(&self) -> &AccountIdOf<Self::T>;
+
 	/// Returns a reference to the account id of the current contract.
 	fn address(&self) -> &AccountIdOf<Self::T>;
 
@@ -294,12 +297,10 @@ where
 	/// The specified `origin` address will be used as `sender` for. The `origin` must be a regular
 	/// account (not a contract).
 	pub fn top_level(origin: T::AccountId, cfg: &'a Config<T>, vm: &'a V, loader: &'a L) -> Self {
-		let source_account = origin;
-		let original_account = source_account.clone();
 		ExecutionContext {
 			parent: None,
 			self_trie_id: None,
-			self_account: source_account,
+			self_account: origin.clone(),
 			overlay: OverlayAccountDb::<T>::new(&DirectAccountDb),
 			depth: 0,
 			deferred: Vec::new(),
@@ -308,7 +309,7 @@ where
 			loader: &loader,
 			timestamp: T::Time::now(),
 			block_number: <system::Module<T>>::block_number(),
-			origin: original_account
+			origin: origin.clone()
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


aliXsed:
@holygits This is my understanding of the task. Please correct me if I'm wrong.

aliXsed:
@holygits Also do we need to add some unit tests to check this?

jordy25519:
Yeah this is it. There should be tests for
- origin is correct at top level call
- origin is correct for nested call

frame/contracts/src/exec.rs
```diff
@@ -293,10 +294,12 @@ where
 	/// The specified `origin` address will be used as `sender` for. The `origin` must be a regular
 	/// account (not a contract).
 	pub fn top_level(origin: T::AccountId, cfg: &'a Config<T>, vm: &'a V, loader: &'a L) -> Self {
+		let source_account = origin;
+		let original_account = source_account.clone();
 		ExecutionContext {
 			parent: None,
 			self_trie_id: None,
-			self_account: origin,
+			self_account: source_account,
```
jordy25519: I think `origin.clone()` would be clearer

frame/contracts/src/exec.rs
```diff
@@ -674,6 +681,7 @@ struct CallContext<'a, 'b: 'a, T: Trait + 'b, V: Vm<T> + 'b, L: Loader<T>> {
 	value_transferred: BalanceOf<T>,
 	timestamp: MomentOf<T>,
 	block_number: T::BlockNumber,
+	origin: T::AccountId, // origin has the same value as caller except for nested calls
```
jordy25519, line 685: Can be a doc comment: ```///  `origin` is the account which authored the original extrinsic leading to the current contract call(s). It does not change on nested calls as with `caller` ```

frame/contracts/src/exec.rs
aliXsed: I wanted to call only one clone, but I agree origin.clone would be clearer.

cowboy-bebug:
For the build issue, we need to resolve this error:
```
   Compiling pallet-elections-phragmen v2.0.0 (/Users/ericlim/project/plug-blockchain-2/frame/elections-phragmen)
   Compiling sp-consensus-babe v2.0.0 (/Users/ericlim/project/plug-blockchain-2/primitives/consensus/babe)
   Compiling pallet-contracts v2.0.0 (/Users/ericlim/project/plug-blockchain-2/frame/contracts)
error[E0407]: method `origin` is not a member of trait `Ext`
   --> /Users/ericlim/project/plug-blockchain-2/frame/contracts/src/exec.rs:763:2
    |
763 |       fn origin(&self) -> &T::AccountId {
    |  _____^
764 | |         &self.origin
765 | |     }
    | |_____^ not a member of trait `Ext`
```

cowboy-bebug (changes requested):
Nice and clean! Just need to add `fn origin()` to the `Ext` trait to resolve build issue:
https://github.com/plugblockchain/plug-blockchain/blob/0966e9d414e018f7a1ab85b6ed145eb78af590a6/frame/contracts/src/exec.rs#L96

aliXsed:
> Nice and clean! Just need to add `fn origin()` to the `Ext` trait to resolve build issue:
> https://github.com/plugblockchain/plug-blockchain/blob/0966e9d414e018f7a1ab85b6ed145eb78af590a6/frame/contracts/src/exec.rs#L96

Thank you Eric! I will do. I created the PR early even before building it as a draft PR to make sure I'm on the right track.

frame/contracts/src/exec.rs
aliXsed: I added the doc to the public interface instead.

cowboy-bebug (changes requested):
Looks good! I think we might want to add one more test for origin() like the tests below:
https://github.com/plugblockchain/plug-blockchain/blob/0966e9d414e018f7a1ab85b6ed145eb78af590a6/frame/contracts/src/wasm/mod.rs#L767
https://github.com/plugblockchain/plug-blockchain/blob/0966e9d414e018f7a1ab85b6ed145eb78af590a6/frame/contracts/src/wasm/mod.rs#L829

cowboy-bebug (approved):
Yep that's it! Nice work!! 👍

PR #39, opened 2019-12-19
MultiAccounting trait

jordy25519:
Introduces a `MultiCurrencyAccounting` trait.

It provides a multi-currency accounting abstraction for generic assets (existing `Currency` trait is a bad name because it conflates the issuance part of the monetary system with the accounting part). Most modules only want the accounting part.

It is intended as a drop-in replacement for `Currency` in contracts and beyond.

Main changes are:
- Add and implement `MultiCurrencyAccounting` trait for generic-asset module
- Change `Imbalance` structs to be "currency aware" so that issuance is managed properly
- Renamed `AssetIdProvider` to `AssetIdAuthority` as these types are backed by blockchain storage in practise
- Added `InherentAssetIdProvider` trait for multi-currency balance types

Other thoughts:
- In general, the definition of `Balance` could change to be currency aware so it is a `(amount, currency_id)`

TODO:
- [x] Happy paths
- [x] Test it defaults to spending currency when `None` is given
- [x] Test error paths are working
  - [x] withdraw 
  - [x] transfer
  - [x] deposit_creating

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,5 @@
 # Changelog
+Track changes made between Plug and it's upstream project `Substrate`.
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
@@ -6,7 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-## [1.0.0] - 2019-10-21
+## Added
+- Add `MultiCurrencyAccounting` trait to support multi currency accounting in contracts module (#39)
+
+## Changed
+- Renamed trait `AssetIdProvider` to `AssetIdAuthority` to reflect it's 'read from chain' behaviour (#39)
+- Make GA imbalance types currency aware so that issuance is managed properly on Drop (#39)
+
+## [1.0.0-rc1] - 2019-10-21
 
 ## Added
 
@@ -41,3 +49,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 `srml/support/src/dispatch.rs`
 - Add `DelegatedDispatchVerifier` check to `decl_module!` expansion. This allows doughnut proofs to be checked when an extrinsic is dispatched using the `<T as system::Trait>::DelegatedDispatchVerifier` impl  
+
+- Renamed binary to `plug` changes made to (`Cargo.toml`s and `Dockerfile` to support this)
+
+### Removed
+
+- The majority of `docs/` is substrate specific and has been removed
+- `README.adoc` is substrate specific and has been removed
--- a/frame/generic-asset/src/impls.rs
+++ b/frame/generic-asset/src/impls.rs
@@ -1,0 +1,574 @@
+// Copyright 2019 Plug New Zealand Ltd.
+// This file is part of Plug.
+
+// Plug is free software: you can redistribute it and/or modify
+// it under the terms of the GNU General Public License as published by
+// the Free Software Foundation, either version 3 of the License, or
+// (at your option) any later version.
+
+// Plug is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+
+// You should have received a copy of the GNU General Public License
+// along with Plug.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Extra trait implementations for the `GenericAsset` module
+
+use crate::{Module, NegativeImbalance, PositiveImbalance, SpendingAssetIdAuthority, Trait};
+use rstd::result;
+use sp_runtime::traits::CheckedSub;
+use support::dispatch::Result;
+use support::{
+	additional_traits::{AssetIdAuthority, MultiCurrencyAccounting},
+	traits::{ExistenceRequirement, Imbalance, SignedImbalance, UpdateBalanceOutcome, WithdrawReasons},
+};
+
+impl<T: Trait> MultiCurrencyAccounting for Module<T> {
+	type AccountId = T::AccountId;
+	type CurrencyId = T::AssetId;
+	type Balance = T::Balance;
+	type DefaultCurrencyId = SpendingAssetIdAuthority<T>;
+	type PositiveImbalance = PositiveImbalance<T>;
+	type NegativeImbalance = NegativeImbalance<T>;
+
+	fn total_balance(who: &T::AccountId, currency: Option<T::AssetId>) -> Self::Balance {
+		<Module<T>>::total_balance(&currency.unwrap_or_else(|| Self::DefaultCurrencyId::asset_id()), who)
+	}
+
+	fn free_balance(who: &T::AccountId, currency: Option<T::AssetId>) -> Self::Balance {
+		<Module<T>>::free_balance(&currency.unwrap_or_else(|| Self::DefaultCurrencyId::asset_id()), who)
+	}
+
+	fn deposit_creating(
+		who: &T::AccountId,
+		currency: Option<T::AssetId>,
+		value: Self::Balance,
+	) -> Self::PositiveImbalance {
+		let asset_id = &currency.unwrap_or_else(|| Self::DefaultCurrencyId::asset_id());
+		let (imbalance, _) =
+			Self::make_free_balance_be(who, currency, <Module<T>>::free_balance(asset_id, who) + value);
+		if let SignedImbalance::Positive(p) = imbalance {
+			p
+		} else {
+			// Impossible, but be defensive.
+			Self::PositiveImbalance::zero()
+		}
+	}
+
+	fn deposit_into_existing(
+		who: &T::AccountId,
+		currency: Option<T::AssetId>,
+		value: Self::Balance,
+	) -> result::Result<Self::PositiveImbalance, &'static str> {
+		// No existential deposit rule and creation fee in GA. `deposit_into_existing` is same with `deposit_creating`.
+		Ok(Self::deposit_creating(who, currency, value))
+	}
+
+	fn ensure_can_withdraw(
+		who: &T::AccountId,
+		currency: Option<T::AssetId>,
+		amount: Self::Balance,
+		reasons: WithdrawReasons,
+		new_balance: Self::Balance,
+	) -> Result {
+		<Module<T>>::ensure_can_withdraw(
+			&currency.unwrap_or_else(|| Self::DefaultCurrencyId::asset_id()),
+			who,
+			amount,
+			reasons,
+			new_balance,
+		)
+	}
+
+	fn make_free_balance_be(
+		who: &T::AccountId,
+		currency: Option<T::AssetId>,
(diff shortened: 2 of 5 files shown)


frame/generic-asset/src/lib.rs
```diff
@@ -864,63 +866,53 @@ pub trait AssetIdProvider {
...
 
 	/// Opaque, move-only struct with private fields that serves as a token denoting that
 	/// funds have been created without any equal and opposite accounting.
 	#[must_use]
-	pub struct PositiveImbalance<T: Subtrait, U: AssetIdProvider<AssetId = T::AssetId>>(
+	#[cfg_attr(test, derive(PartialEq, Debug))]
+	pub struct PositiveImbalance<T: Subtrait>(
 		T::Balance,
-		rstd::marker::PhantomData<U>,
+		T::AssetId,
```
cowboy-bebug: This is very good 👍 👍 🔥

cowboy-bebug (commented):
Generic asset looks so much cleaner with this trait! I'm not sure what the test error is from.

frame/generic-asset/src/impls.rs
```diff
@@ -0,0 +1,551 @@
...
+// You should have received a copy of the GNU General Public License
+// along with Plug.  If not, see <http://www.gnu.org/licenses/>.
+
+//! Extra trait implementations for the `GenericAsset` module
+
+use crate::{Module, NegativeImbalance, PositiveImbalance, SpendingAssetCurrency, Trait};
+use rstd::result;
+use sp_runtime::traits::CheckedSub;
+use support::dispatch::Result;
+use support::{
```
aliXsed, line 23: The formatting of this file looks weird. The tabs are 8 spaces instead of 4 I guess.

frame/support/src/additional_traits.rs
```diff
@@ -83,3 +89,139 @@ impl MaybeDoughnutRef for () {
...
 }
+
+// Note: in the following traits the terms:
+// - 'token' / 'asset' / 'currency' and
+// - 'balance' / 'value' / 'amount'
+// are used interchangeably as they make more sense in certain contexts.
+
+/// An abstraction over the accounting behaviour of a multiple currency system
+/// Currencies in the system are identifiable by a unique `CurrencyId`
+pub trait MultiAccounting<AccountId, CurrencyId> {
```
aliXsed: Do we need to generalize around the currency id? To me, the currency id is more a type than a generic parameter.

aliXsed: Could AccountId be an associated type?

frame/support/src/additional_traits.rs
```diff
@@ -83,3 +89,139 @@ impl MaybeDoughnutRef for () {
...
+
+	/// The default network currency
+	/// When the currency ID is not specified for a `MultiAccounting` method, it will be used
+	/// by default
+	type NetworkCurrency: Currency<AccountId>;
```
aliXsed: It seems to me that NetworkCurrency should be a value (not a type) of the currency id type.

frame/support/src/additional_traits.rs
```diff
@@ -83,3 +89,139 @@ impl MaybeDoughnutRef for () {
...
+}
+
+/// A type which can provide it's currency ID
+/// It is useful in the context of an asset/currency aware balance type
+/// It differs from `AssetIdProvider` in that it is not statically defined
+pub trait CurrencyIdProvider {
+	/// The Currency unique ID type e.g. a `u32`
+	type CurrencyId;
+	/// Return the currency ID inherent to this type
+	fn currency_id(&self) -> Self::CurrencyId;
```
aliXsed: How we are going to ensure that the id would be unique?

frame/support/src/additional_traits.rs
aliXsed: At the very least, I'd say we would need to place a bound on currencyId.

frame/generic-asset/src/impls.rs
```diff
@@ -0,0 +1,551 @@
...
+		additional_traits::MultiAccounting,
+		traits::{
+				Currency, ExistenceRequirement, Imbalance, SignedImbalance, UpdateBalanceOutcome,
+				WithdrawReasons,
+		},
+};
+
+impl<T> MultiAccounting<T::AccountId, T::AssetId> for Module<T>
```
aliXsed: As a type if we go for using object names, we'd better name this type MultiCurrencyAccountant or something similar.

frame/generic-asset/src/impls.rs
```diff
@@ -0,0 +1,551 @@
...
+		fn total_balance(who: &T::AccountId, currency: Option<T::AssetId>) -> Self::Balance {
+				match currency {
+						Some(id) => Self::free_balance(&id, who) + Self::reserved_balance(&id, who),
+						None => Self::NetworkCurrency::total_balance(who),
+				}
+		}
+
+		fn free_balance(who: &T::AccountId, currency: Option<T::AssetId>) -> Self::Balance {
+				match currency {
+						Some(id) => Self::free_balance(&id, who),
```
aliXsed: There seems to be a recursion here swapping the place of the account id and the currency id which I don't understand. Is that really what we want? How this is going to compile when we have only one function called free_balance?

frame/support/src/additional_traits.rs
aliXsed: I would add an associated constant for the network currency id and where ever the currency id is not known I would use that default id. This helps us not to need to have different branches/match arms and duplication of the code for the network currency id.

frame/generic-asset/src/impls.rs
cowboy-bebug: Yeah plug uses hard tab, which is rendered as 8 spaces in github (whereas cennznet uses soft tab).

jordy25519: I actually ran `cargo fmt` on this file but there's no config so I guess it defaulted to this.

frame/support/src/additional_traits.rs
jordy25519: I'll make them both associated

frame/support/src/additional_traits.rs
jordy25519: Tricky part is the spending currency ID can be set dynamically on-chain so I don't think a compile time constant is appropriate

frame/support/src/additional_traits.rs
jordy25519: That's not a requirement for this trait. It's just something which can have it's currency id queried

frame/generic-asset/src/impls.rs
jordy25519: GA can use some clean up to make this more clear. I'd prefer to do that in a follow up. `Self::free_balance` refers to the function defined on the Module itself, it's not recrusive

frame/generic-asset/src/impls.rs
jordy25519: I normally go for trait describes a behaviour not an object. Maybe `MultiCurrencyAccounting`?

frame/support/src/additional_traits.rs
aliXsed: So let's make it a just a function that would return the network currency id.

frame/generic-asset/src/impls.rs
aliXsed: I think MultiCurrencyAccounting is better :)

cowboy-bebug:
Just a reminder to rebase to make tests pass

frame/generic-asset/src/impls.rs
cowboy-bebug: Yeah `MultiCurrencyAccounting` sounds good

frame/support/src/additional_traits.rs
cowboy-bebug: NetworkCurrency is set to be `SpendingAssetCurrency<T>`, which we can get by `spending_asset_id()`

CHANGELOG.md
```diff
@@ -1,12 +1,16 @@
...
 All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
 
-## [1.0.0] - 2019-10-21
+## Added
+- Add `MultiAccounting` trait to support multi currency accounting in contracts module
```
aliXsed: MultiCurrencyAccounting

aliXsed (approved):
Nice Job :)

cowboy-bebug (approved):
Nice thanks!

frame/support/src/additional_traits.rs
```diff
@@ -12,7 +12,7 @@ use sp_runtime::traits::{
 	DoughnutApi, MaybeSerializeDeserialize, SimpleArithmetic, Zero,
 };
 
-/// Perform fee payment for an extrinsic
+/// Perform fee payment for an extsrinsic
```
cowboy-bebug: `extrinsic`

mzxyz (approved):

frame/contracts/src/exec.rs
```diff
@@ -148,6 +149,9 @@ pub trait Ext {
 	/// Returns a reference to the account id of the current contract.
 	fn address(&self) -> &AccountIdOf<Self::T>;
 
+	/// Returns the doughnut of the current contract.
+	fn doughnut(&self) -> Option<&DoughnutOf<Self::T>>;
```
jordy25519: These changes doughnut/contract shouldn't be in this PR. It should already be in `develop` from #44?

frame/generic-asset/src/impls.rs
```diff
@@ -0,0 +1,594 @@
...
+	}
+
+	#[test]
+	fn multi_accounting_withdraw_reserved_funds_should_fail() {
+		let (alice, asset_id, amount) = (1, 16000, 100);
+		ExtBuilder::default()
+			.free_balance((asset_id, alice, amount))
+			.build()
+			.execute_with(|| {
+				// Lock alice's funds
```
jordy25519: `// Reserve alice's funds`

frame/generic-asset/src/impls.rs
```diff
@@ -0,0 +1,594 @@
...
+	fn multi_accounting_deposit_creating_overflow_should_panic() {
+		let (alice, asset_id, amount) = (1, 16000, 100);
+		ExtBuilder::default()
+			.build()
+			.execute_with(|| {
+				let max_value = u64::max_value();
+				let _ = <GenericAsset as MultiCurrencyAccounting>::deposit_creating(
+					&alice, 
+					Some(asset_id), 
+					max_value + amount
```
jordy25519: In this test overflow happens here before the (`::deposit_creating`) is called?
Can remove this test. Also nothing should panic in the runtime, it should return an error

jordy25519 (commented):
TY new tests are good.
The PR history is mixed up with #44 and needs to be resolved before merging

frame/generic-asset/src/impls.rs
mzxyz: Will remove this one

mzxyz:
@holygits  all comments are fixed, if all things looks good, I would squash all the unsigned commits

jordy25519:
LGTM 🚀

cowboy-bebug:
Nice work on this guys 🥇

PR #41, opened 2019-12-22
break the main job into two

cowboy-bebug:
Changes:
- break the main `build` job into `build-bin` and `build-test-and-run`
- use workflow to run them in parallel
- refactor some inline runs as commands
- add 30 min timeout period for running tests


hoani (approved):
LGTM ⛵️

Issue #32, opened 2019-11-21
Initial smart contract permissioning


jordy25519:
Provide initial trait and hooks for smart contract permissioning

- [x] Augment the `DelegatedDispatchVerifier` trait
It should provide functions to handle the following call scenarios:
- runtime calls contract
`verify_runtime_to_contract_dispatch(caller, doughnut, contract address) -> Result`

- contract calls contract
`verify_contract_to_contract_dispatch(caller, doughnut, contract address) -> Result`

- [x] Implement the hook for runtime to contract dispatch (`<T as system::Trait>::DelegatedDispatchVerifier>::verify_runtime_to_contract_call`)
https://github.com/plugblockchain/plug-blockchain/blob/b93e3585eb9bb9f1648a0d155e67832a8aa7f30c/srml/contracts/src/lib.rs#L570-L579

aliXsed:
The following commit completed the task: https://github.com/plugblockchain/plug-blockchain/commit/d5cbf41666a6c7cccbda06d8b7b8d9d8d51747d3

PR #40, opened 2019-12-22
refactor Plug to use the new versioned Doughnut enum type

mzxyz:
- refactor Plug to use the new versioned Doughnut enum type
- creating and tagging a new release branch for `doughnut-rs`  and bump new version `0.3.0`

Updated the original issue, need more time to solve this one

Closes #37

PR #42, opened 2019-12-23
Augment DelegatedDispatchVerifier trait

aliXsed:
Implements step 1 of #32

### DIFF
--- a/frame/executive/src/lib.rs
+++ b/frame/executive/src/lib.rs
@@ -394,6 +394,7 @@ mod tests {
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+		type AccountId = <Self as system::Trait>::AccountId;
 		const DOMAIN: &'static str = "";
 		fn verify_dispatch(
 			_doughnut: &T::Doughnut,
--- a/frame/executive/src/lib.rs
+++ b/frame/executive/src/lib.rs
@@ -394,7 +394,7 @@ mod tests {
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
-		type AccountId = <Self as system::Trait>::AccountId;
+		type AccountId = T::AccountId;
 		const DOMAIN: &'static str = "";
 		fn verify_dispatch(
 			_doughnut: &T::Doughnut,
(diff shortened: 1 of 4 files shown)


frame/executive/src/lib.rs
```diff
@@ -394,6 +394,7 @@ mod tests {
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+		type AccountId = <Self as system::Trait>::AccountId;
```
cowboy-bebug: I think this could be:
```rust
		type AccountId = T::AccountId;
```
because T is bound to system::Trait already so AccountId type can be pulled out from T

frame/executive/tests/doughnut_integration.rs
```diff
@@ -59,6 +59,7 @@ impl_outer_dispatch! {
 
 pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+	type AccountId = <Self as system::Trait>::AccountId;
```
cowboy-bebug: This could be:
```rust
		type AccountId = T::AccountId;
```

cowboy-bebug (commented):
Yep looks good so far. Hope the build issue is not too complicated 😄

frame/executive/src/lib.rs
aliXsed: Yeah I realized that. I'm fixing them at the moment.

cowboy-bebug (approved):
LGTM!

frame/executive/src/lib.rs
```diff
@@ -394,6 +394,7 @@ mod tests {
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+		type AccountId = T::AccountId;
```
jordy25519, line 397: Now we have mixed associated type (AccountId) and generic type (Doughnut). It would be sensible to stick to one or the other

cowboy-bebug: My bad for missing this. It looks like we don't use this associated type anywhere. I think we should remove it to keep it consistent

aliXsed: We have used AccountId as the type of caller and contract addresses. DelegatedDispatchVerifier is defined generically with no knowledge about the type Doughnut or DoughnutRuntime or even generic asset.
A different solution to the one introduced here could be that we define our trait like this:
`pub trait DelegatedDispatchVerifier<Doughnut, AccountId> { ... }`
 But this is then introducing a new generic parameter which is very similar to an associated type and worse because it makes it possible that we implement DelegatedDispatchVerifier for different combinations of Doughnut and AccountId which is not what we would need.
Introducing a dependency between frame/support and frame/system so we bring in generic asset definitions for the additional traits doesn't seem to be a good solution either.

jordy25519: The issue is that we should do all generic parameters or all associated types

`pub trait DelegatedDispatchVerifier<Doughnut, AccountId> { ... }`

or

```
pub trait DelegatedDispatchVerifier {
  type AccountId;
  type Doughnut;
}
```

but not a mix of both as it's confusing.

It will be good to change to all associated after #40 is resolved

PR #44, opened 2020-01-06
add optional doughnut parameter to contract calls

### LINKED ISSUE #33: Implement hook for contract to contract calls
Implement the hook from #32  for contract to contract calls.

- [ ] Add doughnut field to the contract ExecutionContext
https://github.com/plugblockchain/plug-blockchain/blob/b93e3585eb9bb9f1648a0d155e67832a8aa7f30c/srml/contracts/src/exec.rs#L263

- [ ] Expose execution context doughnut via new doughnut() method on the Ext trait
https://github.com/plugblockchain/plug-blockchain/blob/b93e3585eb9bb9f1648a0d155e67832a8aa7f30c/srml/contracts/src/exec.rs#L93

- [ ] Add contract-to-contract hook using `Ext::doughnut()`
https://github.com/plugblockchain/plug-blockchain/blob/b93e3585eb9bb9f1648a0d155e67832a8aa7f30c/srml/contracts/src/wasm/runtime.rs#L385

cowboy-bebug:
Changes:
- [x] ExecutionContext has optional doughnut field
    - added optional doughnut param to methods where needed
    - for tests and node/runtime, None is used as doughnut for now
- [x]  Expose execution context doughnut via new doughnut() method on the Ext trait
- [x] Add contract-to-contract hook using Ext::doughnut()

Closes #33

### DIFF
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -28,6 +28,7 @@ use support::{
 };
 
 pub type AccountIdOf<T> = <T as system::Trait>::AccountId;
+pub type DoughnutOf<T> = <T as system::Trait>::Doughnut;
 pub type CallOf<T> = <T as Trait>::Call;
 pub type MomentOf<T> = <<T as Trait>::Time as Time>::Moment;
 pub type SeedOf<T> = <T as system::Trait>::Hash;
@@ -147,6 +148,9 @@ pub trait Ext {
 
 	/// Returns a reference to the account id of the current contract.
 	fn address(&self) -> &AccountIdOf<Self::T>;
+
+	/// Returns the doughnut of the current contract.
+	fn doughnut(&self) -> Option<DoughnutOf<Self::T>>;
 
 	/// Returns the balance of the current contract.
 	///
@@ -758,6 +762,10 @@ where
 
 	fn address(&self) -> &T::AccountId {
 		&self.ctx.self_account
+	}
+
+	fn doughnut(&self) -> Option<DoughnutOf<T>> {
+		self.ctx.doughnut.clone()
 	}
 
 	fn caller(&self) -> &T::AccountId {
--- a/frame/contracts/src/exec.rs
+++ b/frame/contracts/src/exec.rs
@@ -150,7 +150,7 @@ pub trait Ext {
 	fn address(&self) -> &AccountIdOf<Self::T>;
 
 	/// Returns the doughnut of the current contract.
-	fn doughnut(&self) -> Option<DoughnutOf<Self::T>>;
+	fn doughnut(&self) -> Option<&DoughnutOf<Self::T>>;
 
 	/// Returns the balance of the current contract.
 	///
@@ -288,7 +288,7 @@ pub struct ExecutionContext<'a, T: Trait + 'a, V, L> {
 	pub timestamp: MomentOf<T>,
 	pub block_number: T::BlockNumber,
 	pub origin: T::AccountId,
-	pub doughnut: Option<T::Doughnut>,
+	pub doughnut: Option<&'a T::Doughnut>,
 }
 
 impl<'a, T, E, V, L> ExecutionContext<'a, T, V, L>
@@ -301,7 +301,7 @@ where
 	///
 	/// The specified `origin` address will be used as `sender` for. The `origin` must be a regular
 	/// account (not a contract).
-	pub fn top_level(origin: T::AccountId, cfg: &'a Config<T>, vm: &'a V, loader: &'a L, doughnut: Option<T::Doughnut>) -> Self {
+	pub fn top_level(origin: T::AccountId, cfg: &'a Config<T>, vm: &'a V, loader: &'a L, doughnut: Option<&'a T::Doughnut>) -> Self {
 		ExecutionContext {
 			parent: None,
 			self_trie_id: None,
@@ -315,7 +315,7 @@ where
 			timestamp: T::Time::now(),
 			block_number: <system::Module<T>>::block_number(),
 			origin: origin.clone(),
-			doughnut: doughnut.clone(),
+			doughnut: doughnut,
 		}
 	}
 
@@ -335,7 +335,7 @@ where
 			timestamp: self.timestamp.clone(),
 			block_number: self.block_number.clone(),
 			origin: self.origin.clone(),
-			doughnut: self.doughnut.clone(),
+			doughnut: self.doughnut,
 		}
 	}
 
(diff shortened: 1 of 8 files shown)


frame/contracts/src/wasm/mod.rs
```diff
@@ -903,6 +909,49 @@ mod tests {
...
+		)
+	)
+
+	;; Load a storage value into the scratch buf.
+	(func (export "call")
+		;; assert $ext_scratch_size == 8
+		(call $assert
+			(i32.eq
+				(call $ext_doughnut)
+				(i32.const 0)
```
jordy25519: This is constant `0`, comment above says `1`

frame/contracts/src/wasm/mod.rs
```diff
@@ -903,6 +909,49 @@ mod tests {
 		).unwrap();
 	}
 
+	/// calls `ext_doughnut`, loads the doughnut from the scratch buffer and
```
jordy25519: This test is only checking the return status code: it should assert the contents of the buffer aswell, like with this one: https://github.com/plugblockchain/plug-blockchain/blob/37c0fcfb34923ce5692932b76fe97b5f7cb68580/frame/contracts/src/wasm/mod.rs#L880-L895

frame/contracts/src/wasm/runtime.rs
```diff
@@ -560,6 +560,18 @@ define_env!(Env, <E: Ext>,
 		Ok(())
 	},
 
+	// Stores the doughnut of the current contract into the scratch buffer.
+	// Returns 0 if doughnut is present, otherwise 1.
+	ext_doughnut(ctx) -> u32 => {
```
jordy25519: We don't need this function to be implemented yet, if at all

frame/contracts/src/exec.rs
```diff
@@ -284,6 +288,7 @@ pub struct ExecutionContext<'a, T: Trait + 'a, V, L> {
 	pub timestamp: MomentOf<T>,
 	pub block_number: T::BlockNumber,
 	pub origin: T::AccountId,
+	pub doughnut: Option<T::Doughnut>,
```
jordy25519: We should make this by ref if possible `Option<&T::Doughnut>` and avoid all the clones

jordy25519 (changes requested):
PR looks good in general
- Don't need `ext_doughnut` call to be implemented for this PR but left feedback anyway ;)
- Pass the doughnut by ref if possible (may require change to @alexsednz s work)

frame/contracts/src/wasm/runtime.rs
cowboy-bebug: Thanks! I'll remove it and the related test 😄

frame/contracts/src/exec.rs
```diff
@@ -284,6 +288,7 @@ pub struct ExecutionContext<'a, T: Trait + 'a, V, L> {
 	pub timestamp: MomentOf<T>,
 	pub block_number: T::BlockNumber,
 	pub origin: T::AccountId,
+	pub doughnut: Option<&'a T::Doughnut>,
```
aliXsed, line 291: Wouldn't it be better that we pass the doughnut in through RawOrigin::Signed(issuer, doughnut) as it is for RawOrigin::Delegated for the sake of consistency?

cowboy-bebug: Yeah I wasn't sure which doughnut to use so just used one from `system::Trait`

aliXsed: Yup, if we do so then we wouldn't have two doughnuts passed in directly and indirectly through RawOrigin.

cowboy-bebug: I see what u mean but they are the same thing as `system::Trait` is defined as:
```rust
pub trait Trait: 'static + Eq + Clone {
    /// The aggregated `Origin` type used by dispatchable calls.
    type Origin:
        Into<Result<RawOrigin<Self::AccountId, Self::Doughnut>, Self::Origin>> + 
        From<RawOrigin<Self::AccountId, Self::Doughnut>> +
    ...
}
```
`Self::Doughnut` inside Origin is the `T::Doughnut`

aliXsed: That's why I suggest we should only have one of them. If they could be different then there could be scenarios where we would need to get either of these doughnuts separately and that could justify passing in two references. Does it make sense?

cowboy-bebug: It's more clear and consistent to me that we are using `system::Trait::Doughnut` type, similar to how RawOrigin and DelegatedDispatchVerifier are also using `system::Trait::Doughnut` type. And we are not passing anything just yet, it's just declaring what the type should be for the doughnut field.

frame/contracts/src/wasm/runtime.rs
```diff
@@ -387,6 +388,15 @@ define_env!(Env, <E: Ext>,
 		let value: BalanceOf<<E as Ext>::T> =
 			read_sandbox_memory_as(ctx, value_ptr, value_len)?;
 
+		if let Some(doughnut) = ctx.ext.doughnut() {
+			type DelegatedDispatchVerifierOf<E> = <<E as Ext>::T as system::Trait>::DelegatedDispatchVerifier;
+			DelegatedDispatchVerifierOf::<E>::verify_contract_to_contract_call(
+				&callee,
```
jordy25519: This one should be the `origin` address

frame/contracts/src/wasm/runtime.rs
```diff
@@ -387,6 +388,15 @@ define_env!(Env, <E: Ext>,
 		let value: BalanceOf<<E as Ext>::T> =
 			read_sandbox_memory_as(ctx, value_ptr, value_len)?;
 
+		if let Some(doughnut) = ctx.ext.doughnut() {
+			type DelegatedDispatchVerifierOf<E> = <<E as Ext>::T as system::Trait>::DelegatedDispatchVerifier;
```
jordy25519: Can move the type alias to the top of the file

hoani (approved):
LGTM

jordy25519 (changes requested):
Looks good! Should add a couple integration tests to check `verify_contract_to_contract_call` works correctly. forgot to request this for `runtime_to_contract` in #47

jordy25519 (approved):

jordy25519:
LGTM 🎉 🎉 🎉

frame/contracts/src/lib.rs
```diff
@@ -572,12 +574,13 @@ decl_module! {
 			dest: <T::Lookup as StaticLookup>::Source,
 			#[compact] value: BalanceOf<T>,
 			#[compact] gas_limit: Gas,
-			data: Vec<u8>
+			data: Vec<u8>,
+			doughnut: Option<T::Doughnut>,
```
jordy25519, line 578: so `doughnut` should be extracted from the `origin`. the additional parameter here and in `instantiate` should be removed

cowboy-bebug: Fixed by https://github.com/plugblockchain/plug-blockchain/pull/49

PR #47, opened 2020-01-09
Feature/32 contract permissioning

aliXsed:
Ensure delegated calls are verified by the doughnut.

Also:
- Make it explicit that DelegatedDispatchVerifier of any 'Trait' would be made out of the same doughnut and account id types used in that 'Trait'. 
- Replace "()" as the dummy DelegatedDispatchVerifier by a clear generic type DummyDispatchVerifier which is consistent with the non-dummy case. 
- Moreover make doughnut an associated type rather than a generic parameter for DelegatedDispatchVerifier the same as it is for AccountId.

### DIFF
--- a/frame/contracts/src/tests.rs
+++ b/frame/contracts/src/tests.rs
@@ -114,8 +114,8 @@ impl system::Trait for Test {
 	type AvailableBlockRatio = AvailableBlockRatio;
 	type MaximumBlockLength = MaximumBlockLength;
 	type Version = ();
-        type Doughnut = ();
-        type DelegatedDispatchVerifier = ();
+    type Doughnut = ();
+    type DelegatedDispatchVerifier = ();
 }
 impl balances::Trait for Test {
 	type Balance = u64;
--- a/frame/executive/src/lib.rs
+++ b/frame/executive/src/lib.rs
@@ -394,7 +394,7 @@ mod tests {
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 	impl<T: system::Trait> DelegatedDispatchVerifier for MockDelegatedDispatchVerifier<T> {
-		type Doughnut = T::Doughnut;		
+		type Doughnut = T::Doughnut;
 		type AccountId = T::AccountId;
 		const DOMAIN: &'static str = "";
 		fn verify_dispatch(
--- a/frame/executive/src/lib.rs
+++ b/frame/executive/src/lib.rs
@@ -394,6 +394,7 @@ mod tests {
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
 	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+		type AccountId = <Self as system::Trait>::AccountId;
 		const DOMAIN: &'static str = "";
 		fn verify_dispatch(
 			_doughnut: &T::Doughnut,
--- a/frame/executive/src/lib.rs
+++ b/frame/executive/src/lib.rs
@@ -393,7 +393,8 @@ mod tests {
 
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
-	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+	impl<T: system::Trait> DelegatedDispatchVerifier for MockDelegatedDispatchVerifier<T> {
+		type Doughnut = T::Doughnut;		
 		type AccountId = T::AccountId;
 		const DOMAIN: &'static str = "";
 		fn verify_dispatch(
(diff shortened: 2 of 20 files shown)


aliXsed:
@Reviewers: The process of building and testing the change locally on my computer took so long that I decided to create PR before finishing the process. So the task is not yet done according to the definition of done and there may be build or test issues that I will fix as they appear. However, I reckon it's worth to have the PR at this stage so we can do the review in parallel.

frame/system/src/lib.rs
```diff
@@ -530,6 +530,29 @@ pub fn ensure_signed<OuterOrigin, AccountId, Doughnut>(o: OuterOrigin) -> Result
...
 
+/// Ensure that 'origin' represents a signed or delegated extrinsic. If 'origin' is a delegated one, ensure that doughnut verifies
+/// the issuers's privilage to call dest (destination contract). Return `Ok` with the account id of the issuer who signed the extrinsic
+/// or delegated it, otherwise `Err`.
+pub fn ensure_verified<T: Trait>(
+    origin: T::Origin,
+    dest: &T::AccountId,
+) -> Result<T::AccountId, &'static str> {
+    match origin.into() {
+        // Assuming the delegation proof has been validated, a `RawOrigin::Delegated` should also be considered a valid `RawOrigin::Signed`
```
aliXsed: I'll fix the comment here which's copy-pasted from ensure_signed.

frame/system/src/lib.rs
```diff
@@ -530,6 +530,29 @@ pub fn ensure_signed<OuterOrigin, AccountId, Doughnut>(o: OuterOrigin) -> Result
 	}
 }
 
+/// Ensure that 'origin' represents a signed or delegated extrinsic. If 'origin' is a delegated one, ensure that doughnut verifies
+/// the issuers's privilage to call dest (destination contract). Return `Ok` with the account id of the issuer who signed the extrinsic
+/// or delegated it, otherwise `Err`.
+pub fn ensure_verified<T: Trait>(
```
jordy25519: A name related to contracts would be good :)

jordy25519 (commented):
The associated type clean ups and dummy dispatch are really nice 👍 

An alternative to `ensure_verified` would be to implement something like the following for the AccountId so you can get the `who` from the `origin` without move-ing it
https://github.com/plugblockchain/plug-blockchain/blob/c3fbfd4bba1420f2a54db8bbe276c8b5c97d0d1f/frame/support/src/origin.rs#L195-L206

aliXsed:
> The associated type clean ups and dummy dispatch are really nice 👍
> 
> An alternative to `ensure_verified` would be to implement something like the following for the AccountId so you can get the `who` from the `origin` without move-ing it
> https://github.com/plugblockchain/plug-blockchain/blob/c3fbfd4bba1420f2a54db8bbe276c8b5c97d0d1f/frame/support/src/origin.rs#L195-L206

@holygits Nice alternative. I could even change the doughnut function itself. Thank you for sharing this solution. Anyway, if the current solution is acceptable, I would prefer not changing it.

frame/system/src/lib.rs
aliXsed: What about ensure_verfied_contract_call?

frame/contracts/src/tests.rs
```diff
@@ -114,8 +114,8 @@ impl system::Trait for Test {
 	type AvailableBlockRatio = AvailableBlockRatio;
 	type MaximumBlockLength = MaximumBlockLength;
 	type Version = ();
-        type Doughnut = ();
-        type DelegatedDispatchVerifier = ();
+    type Doughnut = ();
+    type DelegatedDispatchVerifier = ();
```
cowboy-bebug, line 118: Thanks for fixing the format! was going to fix this in my PR

frame/executive/src/lib.rs
```diff
@@ -393,7 +393,8 @@ mod tests {
 
 	// We aren't testing doughnut verification here just return `Ok(())`
 	pub struct MockDelegatedDispatchVerifier<T: system::Trait>(rstd::marker::PhantomData<T>);
-	impl<T: system::Trait> DelegatedDispatchVerifier<T::Doughnut> for MockDelegatedDispatchVerifier<T> {
+	impl<T: system::Trait> DelegatedDispatchVerifier for MockDelegatedDispatchVerifier<T> {
+		type Doughnut = T::Doughnut;
```
cowboy-bebug: It looks like there are some trailing whitespaces in this line. I had to manually put this option in my vscode setting:
```
{
    ...
    "files.trimTrailingWhitespace": true,
    ...
}
```

frame/support/src/additional_traits.rs
```diff
@@ -51,36 +52,94 @@ impl<T, U> ChargeFee<T> for DummyChargeFee<T, U> {
...
-	fn verify_contract_to_contract_dispatch(_caller: &Self::AccountId, _doughnut: &Doughnut, _contract_addr: &Self::AccountId) -> Result<(), &'static str> {
-		Err("Doughnut contract to contract dispatch verification is not implemented for this domain")
+	/// Check the doughnut authorizes a dispatched call to `module` and `method` for this domain
+    fn verify_dispatch(
+        _doughnut: &Self::Doughnut,
+        _module: &str,
+        _method: &str,
+    ) -> Result<(), &'static str> {
+		Err("Doughnut call to module and method verification not implemented for this domain")	
 	}
```
cowboy-bebug: My github might be rendering this incorrectly but does this line need more whitespaces / tab to match indentation?

frame/system/src/lib.rs
cowboy-bebug: `ensure_verified_contract_call` sounds pretty clear to me

cowboy-bebug (approved):
Using `DummyDispatchVerifier` is good and thanks for using the associated types for DelegatedDispatchVerifier trait. I have some comments on minor formatting stuff but other than that LVGTM!
I would like to merge this in first for my PR.

cowboy-bebug (approved):
We can merge after CI passes 😄

PR #48, opened 2020-01-15
Contract to runtime call permissioning

roy-powerfinance:
Added delegated runtime dispatch that supports doughnut

### DIFF
--- a/frame/contracts/src/doughnut_integration.rs
+++ b/frame/contracts/src/doughnut_integration.rs
@@ -84,22 +84,26 @@ impl Get<u64> for BlockGasLimit {
 #[derive(Clone, Eq, PartialEq, Debug)]
 pub struct Test;
 
-#[derive(Clone, Eq, PartialEq, Debug, Encode, Decode)]
+#[derive(Default, Clone, Eq, PartialEq, Debug, Encode, Decode)]
 pub struct MockDoughnut {
+	/// A mock branching point for verify_runtime_to_contract_call, as a doughnut is verified at different level.
+	/// A doughnut is first verified at runtime via Contract::call().
 	runtime_verifiable: bool,
+	/// A mock branching point for verify_contract_to_contract_call, as a doughnut is verified at different level.
+	/// A doughnut is then verified at contract execution via ext_call().
 	contract_verifiable: bool,
 }
 impl MockDoughnut {
-	pub fn new(
-		runtime_verifiable: bool,
-		contract_verifiable: bool,
-	) -> Self {
-		Self {
-			runtime_verifiable,
-			contract_verifiable,
-		}
+	pub fn set_runtime_verifiable(mut self, verifiable: bool) -> Self {
+		self.runtime_verifiable = verifiable;
+		self
+	}
+	pub fn set_contract_verifiable(mut self, verifiable: bool) -> Self {
+		self.contract_verifiable = verifiable;
+		self
 	}
 }
+
 impl PlugDoughnutApi for MockDoughnut {
 	type PublicKey = [u8; 32];
 	type Timestamp = u32;
@@ -454,7 +458,9 @@ const CODE_CALLER_CONTRACT: &str = r#"
 fn contract_to_contract_call_executes_with_verifiable_doughnut() {
 	let (callee_wasm, callee_code_hash) = compile_module::<Test>(CODE_RETURN_WITH_DATA).unwrap();
 	let (caller_wasm, caller_code_hash) = compile_module::<Test>(CODE_CALLER_CONTRACT).unwrap();
-	let verifiable_doughnut = MockDoughnut::new(true, true);
+	let verifiable_doughnut = MockDoughnut::default()
+		.set_runtime_verifiable(true)
+		.set_contract_verifiable(true);
 	let delegated_origin = RawOrigin::from((Some(ALICE), Some(verifiable_doughnut.clone())));
 
 	ExtBuilder::default().existential_deposit(50).build().execute_with(|| {
@@ -514,7 +520,9 @@ fn contract_to_contract_call_returns_error_with_unverifiable_doughnut() {
 
 	// Doughnut is first verified at runtime before execution,
 	// hence runtime_verifiable set to true to bypass the check.
-	let unverifiable_doughnut = MockDoughnut::new(true, false);
+	let unverifiable_doughnut = MockDoughnut::default()
+		.set_runtime_verifiable(true)
+		.set_contract_verifiable(false);
 	let delegated_origin = RawOrigin::from((Some(ALICE), Some(unverifiable_doughnut.clone())));
 
 	ExtBuilder::default().existential_deposit(50).build().execute_with(|| {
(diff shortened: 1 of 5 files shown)


frame/contracts/src/wasm/runtime.rs
```diff
@@ -657,7 +657,14 @@ define_env!(Env, <E: Ext>,
 		};
 		charge_gas(&mut ctx.gas_meter, ctx.schedule, RuntimeToken::ComputedDispatchFee(fee))?;
 
-		ctx.ext.note_dispatch_call(call);
+		match ctx.ext.doughnut() {
```
jordy25519: There should be both a `dispatch_call` and a `dispatch_delegated_call`.
The first means contract wants to do something for itself, the second means contract wants to do something for user (if doughnut permits).
At the moment user can change contract intended behaviour by passing a doughnut

jordy25519 (changes requested):
👍 Needs refactor so that there are two types of dispatches

frame/contracts/src/wasm/runtime.rs
cowboy-bebug: Thanks for spotting this!

cowboy-bebug: Are we introducing duplicated calls by doing this?

frame/contracts/src/wasm/runtime.rs
```diff
@@ -662,6 +662,28 @@ define_env!(Env, <E: Ext>,
...
+		// Charge gas for dispatching this call.
+		let fee = {
+			let balance_fee = <<E as Ext>::T as Trait>::ComputeDispatchFee::compute_dispatch_fee(&call);
+			approx_gas_for_balance(ctx.gas_meter.gas_price(), balance_fee)
+		};
+		charge_gas(&mut ctx.gas_meter, ctx.schedule, RuntimeToken::ComputedDispatchFee(fee))?;
+
+		if let Some(doughnut) = ctx.ext.doughnut() {
+			// Use of intermediate variable to avoid `#[warn(mutable_borrow_reservation_conflict)]`
+			let d = (*doughnut).clone();
```
cowboy-bebug, line 680: I don't think we need to verify doughnut at this stage (or do we?) because it's verified at
- `ext_call()` (for contract-contract calls) and
- `Contract::call()` (for runtime-contract calls)

jordy25519: Correct, no verification to do here. This is a `contract-runtime` call, it will be seen like any ordinary extrinsic with a doughnut and will be verified in the targeted runtime method

frame/contracts/src/wasm/runtime.rs
```diff
@@ -662,6 +662,28 @@ define_env!(Env, <E: Ext>,
 		Ok(())
 	},
 
+	ext_delegated_dispatch_call(ctx, call_ptr: u32, call_len: u32) => {
```
jordy25519, line 667: This is it 👍. Could use a doc comment

jordy25519 (commented):
Add an integration test and then we're good 🍩

cowboy-bebug:
~I can't trigger `verify_contract_to_contract_call` to return error anymore because `Contract::call` checks for `verify_runtime_to_contract` first before actual contract call 😞~

^ The below commit should work around this 😄

frame/contracts/src/doughnut_integration.rs
```diff
@@ -85,12 +86,17 @@ pub struct Test;
 
 #[derive(Clone, Eq, PartialEq, Debug, Encode, Decode)]
 pub struct MockDoughnut {
-	verifiable: bool,
+	runtime_verifiable: bool,
```
jordy25519, line 91: Can doc comment these for the next guy 😉

frame/contracts/src/doughnut_integration.rs
```diff
@@ -441,23 +454,23 @@ const CODE_CALLER_CONTRACT: &str = r#"
 fn contract_to_contract_call_executes_with_verifiable_doughnut() {
 	let (callee_wasm, callee_code_hash) = compile_module::<Test>(CODE_RETURN_WITH_DATA).unwrap();
 	let (caller_wasm, caller_code_hash) = compile_module::<Test>(CODE_CALLER_CONTRACT).unwrap();
-	let verifiable_doughnut = MockDoughnut::new(true);
+	let verifiable_doughnut = MockDoughnut::new(true, true);
```
jordy25519: Could builder pattern this but not important

frame/contracts/src/doughnut_integration.rs
```diff
@@ -521,7 +537,163 @@ fn contract_to_contract_call_returns_error_with_unverifiable_doughnut() {
...
+					event: MetaEvent::contract(RawEvent::DelegatedDispatched(DJANGO, verifiable_doughnut, true)),
+					topics: vec![],
+				}
+			]
+		);
+	});
+}
+
+#[test]
+fn delegated_contract_to_runtime_call_returns_error_with_unverifiable_doughnut() {
```
jordy25519: Name of this test is wrong as this is a test for `verify_runtime_to_contract_call`

frame/contracts/src/wasm/mod.rs
```diff
@@ -1190,6 +1211,44 @@ mod tests {
...
+		)
+	)
+	(func (export "deploy"))
+
+	(data (i32.const 8) "\00\01\2B\00\00\00\00\00\00\00\E5\14\00")
+)
+"#;
+
+	#[test]
+	fn delegated_dispatch_call_works_with_doughnut() {
```
jordy25519, line 1232: Can add another test for when the doughnut is `None` here, it should error

frame/contracts/src/doughnut_integration.rs
```diff
@@ -521,7 +537,163 @@ fn contract_to_contract_call_returns_error_with_unverifiable_doughnut() {
...
+
+	let (wasm, code_hash) = compile_module::<Test>(CODE_DELEGATED_DISPATCH_CALL).unwrap();
+	let verifiable_doughnut = MockDoughnut::new(true, true);
+	let delegated_origin = RawOrigin::from((Some(DJANGO), Some(verifiable_doughnut.clone())));
+
+	ExtBuilder::default().existential_deposit(50).build().execute_with(|| {
+		Balances::deposit_creating(&ALICE, 1_000_000);
+		Balances::deposit_creating(&DJANGO, 1_000_000);
+
+		assert_ok!(Contract::put_code(Origin::signed(ALICE), 100_000, wasm));
```
jordy25519, line 587: I notice some other tests are using `delegated_origin` to `put_code` can you fix them to use Alice like this one (I guess it was left over from when instantiate needed a `Option<Doughnut>`

jordy25519 (changes requested):
Few clean ups, otherwise keen to merge this ASAP

frame/contracts/src/doughnut_integration.rs
cowboy-bebug: Thanks for this comment! It's very clear how it's constructed now! should follow this pattern as much as possible 😄

frame/contracts/src/doughnut_integration.rs
cowboy-bebug: renaming to `delegated_runtime_to_contract_call_returns_error_with_unverifiable_doughnut`

frame/contracts/src/wasm/mod.rs
cowboy-bebug: `delegated_dispatch_call_fails_without_doughnut` is added

frame/contracts/src/doughnut_integration.rs
```diff
@@ -441,23 +458,25 @@ const CODE_CALLER_CONTRACT: &str = r#"
 fn contract_to_contract_call_executes_with_verifiable_doughnut() {
 	let (callee_wasm, callee_code_hash) = compile_module::<Test>(CODE_RETURN_WITH_DATA).unwrap();
 	let (caller_wasm, caller_code_hash) = compile_module::<Test>(CODE_CALLER_CONTRACT).unwrap();
-	let verifiable_doughnut = MockDoughnut::new(true);
+	let verifiable_doughnut = MockDoughnut::default()
+		.set_runtime_verifiable(true)
```
jordy25519, line 462: Too good 🔥

jordy25519 (approved):
🚀 Looks good!

PR #49, opened 2020-01-17
Fix/return doughnut

cowboy-bebug:
Based on the offline discussion with @holygits, `Contract::call()` and `Contract::instantiate()` do not require optional doughnut parameter. Instead, doughnut is pulled from origin, if exists, via ensure_verified_contract_call() method.


frame/system/src/lib.rs
```diff
@@ -534,21 +534,25 @@ pub fn ensure_signed<OuterOrigin, AccountId, Doughnut>(o: OuterOrigin) -> Result
...
-                T::DelegatedDispatchVerifier::verify_runtime_to_contract_call(&t, &doughnut, dest)
-            {
-                Err(msg)
-            } else {
-                Ok(t)
-            }
-        }
-        _ => Err("bad origin: expected to be a signed origin"),
+	origin: T::Origin,
+	dest: Option<&T::AccountId>,
```
cowboy-bebug: `dest` became optional to bypass `verify_runtime_to_contract_call` check.

frame/contracts/src/lib.rs
```diff
@@ -601,9 +600,8 @@ decl_module! {
 			#[compact] gas_limit: Gas,
 			code_hash: CodeHash<T>,
 			data: Vec<u8>,
-			doughnut: Option<T::Doughnut>,
 		) -> Result {
-			let origin = ensure_signed(origin)?;
+			let (origin, doughnut) = ensure_verified_contract_call::<T>(origin, None)?;
```
jordy25519: No need to call this here, just pass `None` doughnut to `execute_wasm`

frame/system/src/lib.rs
jordy25519: Can undo this since it's not needed in `pub fn instantiate`.
var naming can be improved from what it was though `dest` -> `contract`, `account` -> `caller` or `who` is used regularly throughout the codebase

jordy25519 (commented):
Thanks for fixing this. Some things to fix then LGTM

jordy25519 (approved):
👍 nice work

PR #51, opened 2020-01-23
Merge/upstream

cowboy-bebug:
## Upstream Changes

Pulled and merged changes up to https://github.com/paritytech/substrate/commit/cf020addf0e2d9d5fdaa7a34aa83b380f49471bb.

### Changed:
- Crate aliasing in cargo files is removed, and crates are imported as is
    - some crates are still aliased in `use` statements
- `frame_system::Trait` has added ModuleToIndex type (generated by `construct_runtime!`)
- `pallet_balances::Trait` has added OnReapAccount type (to handle accounts fully reaped from the system)
- `Result<(), &'static str>`  -> mapped to `DispatchError::Other("error msg")` type
- `Result<(), DispatchError>` -> mapped to `DispatchError::Module { index: 0, error: 0, message: Some("err msg") }` type
    - Note: requires using decl_error! macro to create module specific errors (index is generated from the order of entry)
- refactoring here and there to reduce complexity of the code base (less lines and more modular)

### Added crates:
- `network-gossip`: wraps around an implementation of the `Network` crate and provides gossiping capabilities on top of it
- `recovery`: for users to recovery their accounts if private key / auth mechanism is lost
- `society`: an economic game module which incentivizes users to participate and maintain a membership society (a pot is created, passed around and a member can claim it via voting and stuff)
- a few test crates to test existing runtime modules


cowboy-bebug:
Reopened at https://github.com/plugblockchain/plug-blockchain/pull/55

PR #54, opened 2020-01-24
Decouple gas handling from gas payment

### LINKED ISSUE #53: Make filling and refunding gas configurable
https://centralitydev.atlassian.net/browse/BT-124
https://centralitydev.atlassian.net/browse/BT-131

aliXsed:
Closes #53 
Create a separate trait for gas handling so we can customize its behaviour in the case of multi-currency gas payment in which we don't want the user to be charged upfront.


frame/contracts/src/gas.rs
```diff
@@ -194,6 +194,34 @@ impl<T: Trait> GasMeter<T> {
...
+///
+/// This trait allow customization on how the `GasMeter` will be filled and emptied.
+///
+/// If your economic model does not require upfront payment for the gas to be filled,
+/// you can impl this trait with your own business logic
+///
+pub trait GasHandling<T: Trait> {
+	/// This function fills the gas meter with a certain amount of gas
+	/// Default behaviour will withdraw currency from the user's balance upfront to pay for the gas
+	fn fill_gas(transactor: &T::AccountId, gas_limit: Gas) -> Result<(GasMeter<T>, NegativeImbalanceOf<T>), &'static str>
```
jordy25519: The imbalance result from `fill_gas` should be made into an `Option` as it implies some imbalance was generated (which may not be the case).

The `imbalance` parameter for `return_unused_gas` should also be an `Option` as it seems to be used as a way to pass the imbalance from the buy step into the refund step.

Passing the imbalance, instead of dropping immediately, appears to be an optimization.
If it was not passed between the methods then there would be two updates to the currency's issuance. One when gas is bought (negative imbalance) and another when gas is refunded (positive imbalance).
At the moment the two get merged into one when the `imbalance.offset(refund)` is called. This means there is only one update to the currency's issuance, instead of two.

jordy25519 (changes requested):
Yeah this looks good 👍 
We need to tidy up the imbalance that is being returned and passed around
It should be made into an `Option` since generating an imbalance is optional (only happens if we mint or burn currency) but there are two ways we could implement it
1) Make `fill_gas` return `Option<NegativeImbalance>` and `return_unused_gas` receive `Option<NegativeImbalance>`
**OR**
2) Make `imbalance` into a field on the gas meter itself so this imbalance "state" does not need to be passed around which seems cleaner. It should still be an `Option<NegativeImbalance>` though, maybe named like `pending_imbalance`

Last note would be `GasHandling` -> `GasHandler` to fit the naming convention

frame/contracts/src/gas.rs
```diff
@@ -194,6 +194,34 @@ impl<T: Trait> GasMeter<T> {
...
+	{
+		buy_gas::<T>(transactor, gas_limit)
+	}
+
+	/// This function empties the remaining gas in the gas meter
+	/// Default behaviour will deposit currency into the user's balance to refund un-used gas
+	fn return_unused_gas(
```
hoani: Because we don't require that gas is actually returned here, I suggest naming this `handle_unused_gas`

frame/contracts/src/gas.rs
hoani: Are we sure we want to include imbalance logic in the contracts module?
I think gas belongs here, but imbalance logic is currency specific

frame/contracts/src/gas.rs
hoani: ^ looking at the description, `empty_remaining_gas` is a good name too

frame/contracts/src/gas.rs
jordy25519: The imbalance logic is already tied into the contracts module for other currency related stuff (e.g. a contract transferring assets). We could remove it from the "gas" part of the module, but removing from the whole thing is a much larger issuer

cowboy-bebug: Instead of changing the return type to be Option, can we do something like this in GasHandling::buy_gas():
```rust
buy_gas::<T>(transactor, gas_limit)
	.map(|(gas_meter, imbalance)|) {
		if imbalance.is_zero() {
			(gas_meter, 0.into())
		}
	})
```
And propagate 0 imbalance throughout execute_wasm() and put_code()

aliXsed: @royyang-centrality @hoani and I discussed different options and decided to get rid of this coupling altogether. As @holygits pointed out passing imbalance was introduced as an optimization measure so during a contract execution they call on_unbalance only once after refunding. This trivial optimization would cost us an ugly coupling between gas payment and gas handling if we insist to keep it. The alternative solution that we have taken is we settle any imbalance whenever it is incurred at its own time. So a payment or a deposit will deal with imbalance internally and appear complete to outside.

frame/contracts/src/gas.rs
```diff
@@ -239,8 +267,9 @@ pub fn refund_unused_gas<T: Trait>(
 	// Refund gas left by the price it was bought at.
 	let refund = gas_meter.gas_price * gas_left.unique_saturated_into();
 	let refund_imbalance = T::Currency::deposit_creating(transactor, refund);
-	if let Ok(imbalance) = imbalance.offset(refund_imbalance) {
-		T::GasPayment::on_unbalanced(imbalance);
+
+	if let Ok(imbalance) = NegativeImbalanceOf::<T>::zero().offset(refund_imbalance) {
```
jordy25519: can we just use `refund_imbalance`, without offset?

frame/contracts/src/lib.rs
```diff
@@ -357,6 +357,8 @@ pub trait Trait: system::Trait {
 	/// Handler for the unbalanced reduction when making a gas payment.
 	type GasPayment: OnUnbalanced<NegativeImbalanceOf<Self>>;
 
+	type GasHandler: GasHandler<Self>;
```
jordy25519, line 374: Add a doc comnent 😄

frame/contracts/src/gas.rs
jordy25519: Much cleaner now. I think the point of the optimization was so that there would only be one write to storage

frame/contracts/src/gas.rs
aliXsed: No. refund_imbalance is PositiveImbalance and on_unbalanced takes NegativeImbalance.

jordy25519: Ok so is this code passing the refund as a negative imbalance? it should just be zero negative imbalance if that is the case

jordy25519: Actually, I think we can remove this altogether, as it was intended to drop the "net" negative imbalance across buy and refund steps. But we already drop the negative imbalance in buy step now

aliXsed: Would you please elaborate on your point a bit? Previously we would pass
`neg_imb - pos_imb`
Now because we have settled "neg_imb" at the payment site, so here neg_imb is zero. That means we should pass:
`zero - pos_imb`

jordy25519: Won't this create a negative imbalance? `zero - pos_imb`

aliXsed: Even if on_unbalance is going to ignore what we are passing in the second step, I would say let's keep it, so in any possible future changes we could keep accounting simple by maintaining symmetry.

aliXsed: What I know about `zero - pos_imb` is that it's consistent with the previous logic here.

aliXsed: By reading the code I guess it would result in Err and thus will be ignored

aliXsed: Anyway, if you think the logic of settling imbalance would not change substantially, removing it altogether seems not to be wrong.

jordy25519: I think this will always be a noop. refund will be 0 or positive, if its 0 it will make another 0, if its positive it will make a `Err<PositiveImbalance>` and miss the let binding. SO makes sense to remove, since this code was here based on the assumption a "net" negative imbalance would be passed in

frame/contracts/src/gas.rs
```diff
@@ -194,14 +194,41 @@ impl<T: Trait> GasMeter<T> {
...
+///
+/// This trait allow customization on how the `GasMeter` will be filled and emptied.
+///
+/// If your economic model does not require upfront payment for the gas to be filled,
+/// you can impl this trait with your own business logic
+///
+pub trait GasHandler<T: Trait> {
+	/// This function fills the gas meter with a certain amount of gas
+	/// Default behaviour will withdraw currency from the user's balance upfront to pay for the gas
+	fn fill_gas(transactor: &T::AccountId, gas_limit: Gas) -> Result<GasMeter<T>, &'static str>
```
mzxyz: Awesome work!  A small case I would recommend is to add associated `type Error` instead use `&'static str` directly, something like this: `type Error: From<Error> + std::fmt::Debug + 'static;` , For example: it would be more flexible if we want to support multi errors with Enum

cowboy-bebug: I think all `Result<(), &'static str>` changed to `Result<(), DispatchError>` or `DispatchResult` in the latest merge as well

frame/contracts/src/gas.rs
cowboy-bebug: Yep `deposit_creating()` always returns +ve or zero PositiveImbalance. And to make on_unbalanced work for PositiveImbalance, the below may have to change:

https://github.com/plugblockchain/plug-blockchain/blob/55aa532059755cc37141949fc6f428a305b07c44/frame/contracts/src/lib.rs#L371

frame/contracts/src/gas.rs
mzxyz: You are right https://github.com/plugblockchain/plug-blockchain/blob/55aa532059755cc37141949fc6f428a305b07c44/primitives/runtime/src/lib.rs#L359  Using `Result<(), DispatchError>` or `DispatchResult` would be the best practice for now

aliXsed: Done

jordy25519 (approved):
Very clean now, nice job! LGTM (pending CI passes and an integration test, could be in follow up PR)

frame/contracts/src/lib.rs
```diff
@@ -736,16 +738,12 @@ impl<T: Trait> Module<T> {
 		doughnut: Option<T::Doughnut>,
 		func: impl FnOnce(&mut ExecutionContext<T, WasmVm, WasmLoader>, &mut GasMeter<T>) -> ExecResult
 	) -> ExecResult {
-		// Pay for the gas upfront.
-		//
-		// NOTE: it is very important to avoid any state changes before
-		// paying for the gas.
```
cowboy-bebug, line 742: Could we keep this comment (for the next person who works on this bit of code). Something like this:
```rust
// Fill up the gas meter upfront. Default behaviour is to pay for the gas upfront.
//
// NOTE: it is very important to avoid any state changes before
// paying for the gas.
```

frame/contracts/src/lib.rs
```diff
@@ -759,11 +757,7 @@ impl<T: Trait> Module<T> {
 			DirectAccountDb.commit(ctx.overlay.into_change_set());
 		}
 
-		// Refund cost of the unused gas.
-		//
-		// NOTE: This should go after the commit to the storage, since the storage changes
-		// can alter the balance of the caller.
```
cowboy-bebug: Can we also keep this comment and modify it to apply to our change:
```rust
// Handle unused gas. Default behaviour is to refund cost of the unused gas.
//
// NOTE: This should go after the commit to the storage, since the storage changes
// can alter the balance of the caller.
```

frame/contracts/src/lib.rs
```diff
@@ -370,6 +370,9 @@ pub trait Trait: frame_system::Trait {
 	/// Handler for the unbalanced reduction when making a gas payment.
 	type GasPayment: OnUnbalanced<NegativeImbalanceOf<Self>>;
 
+	/// Handler for filling and emptying the gas tank for a contract.
```
cowboy-bebug: `/// Handler for filling and emptying the gas meter for a contract.` <- just to avoid any confusion

frame/contracts/src/gas.rs
```diff
@@ -240,9 +268,6 @@ pub fn refund_unused_gas<T: Trait>(
 	// Refund gas left by the price it was bought at.
 	let refund = gas_meter.gas_price * gas_left.unique_saturated_into();
 	let refund_imbalance = T::Currency::deposit_creating(transactor, refund);
```
cowboy-bebug: Can we make it `_refund_imbalance` to avoid `unused_variables` warning

cowboy-bebug (changes requested):
Yep looks good! 🎉 
I got minor changes requested around comments. And if you'd like `NegativeImbalanceOf` import in gas.rs could be removed as we don't use it anymore.

Also a reminder to rebase to the latest develop and update `CHANGELOG.md` with things that this PR brings in 😄

frame/contracts/src/lib.rs
aliXsed: Meter is a gauge counting your consumption. So technically we don't fill the meter.

frame/contracts/src/lib.rs
aliXsed: I am not very pro implementation comments unless you are doing something that may look wrong or has hidden implications. Ideally, It's your code that should be readable. In this case, "we are filling the gas, which is not hard to see"

frame/contracts/src/lib.rs
aliXsed: Similarly in this case. We, instead, have added nice documentation comments to the trait explaining what we do.

frame/contracts/src/lib.rs
cowboy-bebug: We do fill and empty the meter according to `GasHandler` doc comments

aliXsed: Yeah, you are right. That comment is from @royyang-centrality. On a second look, I'd say it's GasMeter that I would rather be called GasTank. But I don't see that much value in refactoring it. So alright. I will change the comment here.

frame/contracts/src/lib.rs
mzxyz: For me, I prfer to keep this comment

mzxyz (approved):

cowboy-bebug (approved):

hoani (approved):

PR #55, opened 2020-01-27
Merge latest upstream/master bc85d35

cowboy-bebug:
## Upstream Changes

Pulled and merged changes up to https://github.com/paritytech/substrate/commit/bc85d352681018e5b19369b758c013a52c5c3032.

### Changed:
- Crate aliasing in cargo files is removed, and crates are imported as is
    - some crates are still aliased in `use` statements
- `frame_system::Trait` has added ModuleToIndex type (generated by `construct_runtime!`)
- `pallet_balances::Trait` has added OnReapAccount type (to handle accounts fully reaped from the system)
- `Result<(), &'static str>`  -> mapped to `DispatchError::Other("error msg")` type
- `Result<(), DispatchError>` -> mapped to `DispatchError::Module { index: 0, error: 0, message: Some("err msg") }` type
    - Note: requires using decl_error! macro to create module specific errors (index is generated from the order of entry)
- refactoring here and there to reduce complexity of the code base (less lines and more modular)

### Added crates:
- `network-gossip`: wraps around an implementation of the `Network` crate and provides gossiping capabilities on top of it
- `recovery`: for users to recovery their accounts if private key / auth mechanism is lost
- `society`: an economic game module which incentivizes users to participate and maintain a membership society (a pot is created, passed around and a member can claim it via voting and stuff)
- a few test crates to test existing runtime modules


bin/node/executor/tests/basic.rs
```diff
@@ -353,7 +353,7 @@ fn full_native_block_import_works() {
 			},
 			EventRecord {
 				phase: Phase::ApplyExtrinsic(1),
-				event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984800000000)),
+				event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1992800000000)),
```
cowboy-bebug, line 358: The values here had to be changed to make the tests pass. I couldn't figure out why

bin/node/executor/tests/basic.rs
```diff
@@ -406,7 +406,7 @@ fn full_native_block_import_works() {
 			},
 			EventRecord {
 				phase: Phase::ApplyExtrinsic(1),
-				event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984788199392)),
+				event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1992788119712)),
```
cowboy-bebug, line 411: This value had to be changed as well to pass tests

bin/node/executor/tests/basic.rs
```diff
@@ -430,7 +430,7 @@ fn full_native_block_import_works() {
 			},
 			EventRecord {
 				phase: Phase::ApplyExtrinsic(2),
-				event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1984788199392)),
+				event: Event::pallet_treasury(pallet_treasury::RawEvent::Deposit(1992788119712)),
```
cowboy-bebug, line 435: This value had to be changed as well to pass tests again

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -173,7 +173,7 @@ fn submitted_transaction_should_be_valid() {
 		let res = Executive::validate_transaction(extrinsic);
 
 		assert_eq!(res.unwrap(), ValidTransaction {
-			priority: 2_411_002_000_000,
+			priority: 2_421_002_000_000,
```
cowboy-bebug, line 176: This value had to be changed as well to pass tests

frame/society/src/tests.rs
```diff
@@ -128,7 +128,7 @@ fn bidding_works() {
 		// Candidates become members after a period rotation
 		assert_eq!(Society::members(), vec![10, 30, 40]);
 		// Pot is increased by 1000, but pays out 700 to the members
-		assert_eq!(Balances::free_balance(Society::account_id()), 9_300);
+		assert_eq!(Balances::free_balance(Society::account_id()), 10_000);
```
cowboy-bebug: pay out is not working so the value had to be changed. I'm not sure if it's due to  changing u128 -> u64 for account type in this crate.

bin/node/executor/tests/basic.rs
jordy25519: treasury deposit should be equivalent to the transaction fees paid. It will be related to some changes to the extrinsic format e.g. we add `Option<PlugDoughnut>` to extrinsics which changes the per byte fee.
See `- transfer_fee` here https://github.com/plugblockchain/plug-blockchain/blob/c93eafe3821a5e4975e741fc39b90a5c88ef8d92/bin/node/executor/tests/basic.rs#L343.
Would be nice to compare with substrate to check the fee changes are as expected

frame/society/src/tests.rs
cowboy-bebug: Using `type DelegatedDispatchVerifier = ()` expected associated AccountId to be u64, Using DummyDispatchVerifier worked 😄

bin/node/executor/tests/submit_transaction.rs
cowboy-bebug: The value is bumped up by this value:
https://github.com/plugblockchain/plug-blockchain/blob/13fd2d78e0cd7fdd62dba4add9130eb74f049f2d/bin/node/runtime/src/lib.rs#L181
(because we added a byte of doughnut in the extrinsic payload)

So it's good!

bin/node/executor/tests/basic.rs
cowboy-bebug: Checked that `transfer_fee` is indeed increased by `TransactionByteFee` const because we use extra byte for doughnut in the extrinsic payload (set `WeightFeeCoefficient = 1` to test it).

cowboy-bebug: I'm not sure how else to check this but I did side-by-side comparison with different `TransactionByteFee` values (and WeightFeeCoefficient = 1) and the difference in value is 0.8 * TransactionByteFee as 80% goes to treasury and 20% goes to the block author.

jordy25519 (approved):
Massive PR, nice work 🚀 
Add some proof comments for the fee amount changes in tests and then LGTM!

bin/node/executor/tests/basic.rs
```diff
@@ -54,6 +54,8 @@ use self::common::{*, sign};
 pub const BLOATY_CODE: &[u8] = node_runtime::WASM_BINARY_BLOATY;
 
 /// Default transfer fee
+// NOTE: Transfer fee increased by 1 byte * TransactionByteFee as we include Option<Doughnut> in SignedExtra.
+//       Option always takes up one byte in extrinsic payload
```
cowboy-bebug, line 58: A note on the transfer fee change is added here

mzxyz:
Awesome work

PR #60, opened 2020-02-02
Chores/merge-upstream

aliXsed:
Merge substrate/master from git@github.com:paritytech/substrate.git commit #de2ffd9


hoani:
@alexsednz to make reviewing a little easier, do you have a list of files which had merge conflicts?

aliXsed:
> @alexsednz to make reviewing a little easier, do you have a list of files which had merge conflicts? Here is the list of files with conflicts:

        Cargo.lock
        bin/node-template/runtime/src/template.rs
        bin/node/cli/Cargo.toml
        bin/node/cli/bin/main.rs
        bin/node/runtime/Cargo.toml
        bin/node/runtime/src/lib.rs
        frame/balances/src/lib.rs
        frame/balances/src/tests.rs
        frame/contracts/rpc/src/lib.rs
        frame/contracts/src/lib.rs
        frame/contracts/src/tests.rs
        frame/executive/src/lib.rs
        frame/generic-asset/src/lib.rs
        frame/identity/src/lib.rs
        frame/support/src/metadata.rs

And here is the list of those that I needed to touch to fix build/test issues:

        client/cli/src/node_key.rs
        client/network/src/behaviour.rs
        frame/authority-discovery/src/lib.rs
        frame/contracts/src/doughnut_integration.rs
        frame/contracts/src/gas_tests.rs
        frame/executive/tests/doughnut_integration.rs
        frame/generic-asset/src/impls.rs
        frame/staking/src/multi_token_economy_tests.rs
        frame/vesting/src/lib.rs
        prml/attestation/src/lib.rs

bin/node/runtime/src/lib.rs
```diff
@@ -77,11 +76,11 @@ pub const VERSION: RuntimeVersion = RuntimeVersion {
 	impl_name: create_runtime_str!("plug-node"),
 	authoring_version: 10,
 	// Per convention: if the runtime behavior changes, increment spec_version
-	// and set impl_version to equal spec_version. If only runtime
+	// and set impl_version to 0. If only runtime
 	// implementation changes and behavior does not, then leave spec_version as
 	// is and increment impl_version.
-	spec_version: 205,
```
hoani, line 83: I think at most our spec version should increase by one with a merge. We are not going to stay up to date with substrate spec versions.

hoani (approved):
LGTM, made a small comment about spec_version, we should probably just increment this by one on merges with substrate

.circleci/config.yml
```diff
@@ -52,12 +52,12 @@ commands:
             - target/debug/build
             - target/debug/deps
             - target/debug/incremental
-          key: target-cache-{{ arch }}-{{ checksum ".circleci/RUST_VERSION" }}-{{ .Environment.CIRCLE_BRANCH }}-{{ epoch }}
+          key: target-cache-{{ arch }}-{{ .Environment.RUST_VERSION }}-{{ .Environment.CIRCLE_BRANCH }}-{{ epoch }}
```
cowboy-bebug, line 55: 👍

cowboy-bebug (approved):
LGTM nice work!!

.circleci/config.yml
```diff
@@ -102,7 +102,7 @@ jobs:
     resource_class: large
     environment:
       BASH_ENV: ~/.cargo/env
-      RUST_VERSION: stable-2019-12-19
+      RUST_VERSION: 1.41.0
```
cowboy-bebug, line 105: if u want to keep the same format as the previous, can try `stable-2020-01-30`

aliXsed: We tried stable-2020-01-27 which failed at `rustup default`. Where did you get stable-2020-01-30 from?
Anyhow, if you are okay too, I believe 1.41.0 is clearer plus we can always simply get it from the rustc version.

bin/node/runtime/src/lib.rs
aliXsed: Yeah, seems like we need to have our own version and change it upon either:

- A merge from upstream that has changed the spec_version
- A change in the plug itself when it affects the spec

I guess we may need to detail a process for it. However, if it's the only change, I would suggest we let the current merge happen and do what we have decided in a separate PR.

.circleci/config.yml
cowboy-bebug: I got the date from rust release page: https://github.com/rust-lang/rust/blob/master/RELEASES.md#version-1410-2020-01-30

jordy25519 (approved):

PR #68, opened 2020-02-24
Uncouple emptying gas_meter from updating the gas spent for the block

### LINKED ISSUE #67: Uncouple handling block gas spent from emptying gas meter
Updating block gas spent should be done independently from emptying gas meter, so if the latter function is overridden, there would be no need to deal with block gas spent too. This will also eliminate the need for exposing the block gas handling function.

aliXsed:
Closes #67

### DIFF
--- a/frame/contracts/src/gas.rs
+++ b/frame/contracts/src/gas.rs
@@ -212,7 +212,7 @@ pub trait GasHandler<T: Trait> {
 	/// This function should be called when the contract has stopped consuming gas and the gas_meter
 	/// is ready to be read. This will update the gas spent for the block and then empties the
 	/// unused gas.
-	fn update(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
+	fn finish(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
 		// Increase total spent gas.
 		// This cannot overflow, since `gas_spent` is never greater than `block_gas_limit`, which
 		// also has Gas type.
--- a/frame/contracts/src/gas.rs
+++ b/frame/contracts/src/gas.rs
@@ -14,14 +14,13 @@
 // You should have received a copy of the GNU General Public License
 // along with Substrate. If not, see <http://www.gnu.org/licenses/>.
 
-use crate::{GasSpent, Module, Trait, BalanceOf};
+use crate::{Module, Trait, BalanceOf};
 use sp_std::convert::TryFrom;
 use sp_runtime::traits::{
 	CheckedMul, Zero, SaturatedConversion, AtLeast32Bit, UniqueSaturatedInto,
 };
 use frame_support::{
-	traits::{Currency, ExistenceRequirement, OnUnbalanced, WithdrawReason}, StorageValue,
-	dispatch::DispatchError,
+	traits::{Currency, ExistenceRequirement, OnUnbalanced, WithdrawReason},	dispatch::DispatchError,
 };
 
 #[cfg(test)]
@@ -207,18 +206,6 @@ pub trait GasHandler<T: Trait> {
 	fn fill_gas(transactor: &T::AccountId, gas_limit: Gas) -> Result<GasMeter<T>, DispatchError>
 	{
 		buy_gas::<T>(transactor, gas_limit)
-	}
-
-	/// This function should be called when the contract has stopped consuming gas and the gas_meter
-	/// is ready to be read. This will update the gas spent for the block and then empties the
-	/// unused gas.
-	fn finish(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
-		// Increase total spent gas.
-		// This cannot overflow, since `gas_spent` is never greater than `block_gas_limit`, which
-		// also has Gas type.
-		GasSpent::mutate(|block_gas_spent| *block_gas_spent += gas_meter.spent());
-
-		Self::empty_unused_gas(transactor, gas_meter);
 	}
 
 	/// This function empties the remaining gas in the gas meter
--- a/frame/contracts/src/gas.rs
+++ b/frame/contracts/src/gas.rs
@@ -209,6 +209,18 @@ pub trait GasHandler<T: Trait> {
 		buy_gas::<T>(transactor, gas_limit)
 	}
 
+	/// This function should be called when the contract has stopped consuming gas and the gas_meter
+	/// is ready to be read. This will update the gas spent for the block and then empties the
+	/// unused gas.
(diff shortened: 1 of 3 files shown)


frame/contracts/src/gas.rs
```diff
@@ -209,6 +209,18 @@ pub trait GasHandler<T: Trait> {
...
+	/// This function should be called when the contract has stopped consuming gas and the gas_meter
+	/// is ready to be read. This will update the gas spent for the block and then empties the
+	/// unused gas.
+	fn update(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
+		// Increase total spent gas.
+		// This cannot overflow, since `gas_spent` is never greater than `block_gas_limit`, which
+		// also has Gas type.
+		GasSpent::mutate(|block_gas_spent| *block_gas_spent += gas_meter.spent());
+
+		Self::empty_unused_gas(transactor, gas_meter);
```
aliXsed: I'm uncertain whether this would actually call the overridden function in the places that we have implemented this trait. I will do a test soon.

aliXsed: My test showed this should work. So we can go ahead.

frame/contracts/src/gas.rs
```diff
@@ -257,14 +269,8 @@ pub fn refund_unused_gas<T: Trait>(
...
 	gas_meter: GasMeter<T>,
 ) {
-	let gas_spent = gas_meter.spent();
 	let gas_left = gas_meter.gas_left();
 
-	// Increase total spent gas.
-	// This cannot overflow, since `gas_spent` is never greater than `block_gas_limit`, which
-	// also has Gas type.
-	GasSpent::mutate(|block_gas_spent| *block_gas_spent += gas_spent);
-
```
cowboy-bebug, line 267: Do we need to call the update method here to ensure the default behaviour here (please ignore me if we don't need to)?
```rust
Self::update(transactor, gas_meter);
```

cowboy-bebug:
~I think the build errors are coming from the way we use the latest nightly in our circleci config. Let me fix it in another PR~ Actually the latest nightly probably fixes this.

jordy25519:
related: https://github.com/rustwasm/wasm-bindgen/issues/2009

frame/contracts/src/gas.rs
jordy25519: Can you add a specific unit test to see a custom impl updates gas_spent

frame/contracts/src/gas.rs
```diff
@@ -209,6 +209,18 @@ pub trait GasHandler<T: Trait> {
 		buy_gas::<T>(transactor, gas_limit)
 	}
 
+	/// This function should be called when the contract has stopped consuming gas and the gas_meter
+	/// is ready to be read. This will update the gas spent for the block and then empties the
+	/// unused gas.
+	fn update(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
```
jordy25519: a more specific name would be good here

frame/contracts/src/gas.rs
aliXsed: It's now the other way around. The update method's default is to call this function. So the default behaviour would remain as it was before.

frame/contracts/src/gas.rs
aliXsed: I made a change so customized_fill_gas_does_not_charge_the_user, if passed, would prove the right empty_unused_gas is called.

frame/contracts/src/gas.rs
aliXsed: Done.

frame/contracts/src/gas.rs
```diff
@@ -209,6 +209,18 @@ pub trait GasHandler<T: Trait> {
 		buy_gas::<T>(transactor, gas_limit)
 	}
 
+	/// This function should be called when the contract has stopped consuming gas and the gas_meter
+	/// is ready to be read. This will update the gas spent for the block and then empties the
+	/// unused gas.
+	fn finish(transactor: &T::AccountId, gas_meter: GasMeter<T>) {
```
hoani: I don't think that `finish` is more descriptive than `update`, but I will let @holygits decide since it was his concern in the first place.

frame/contracts/src/gas.rs
```diff
@@ -209,6 +209,18 @@ pub trait GasHandler<T: Trait> {
...
+		// Increase total spent gas.
+		// This cannot overflow, since `gas_spent` is never greater than `block_gas_limit`, which
+		// also has Gas type.
+		GasSpent::mutate(|block_gas_spent| *block_gas_spent += gas_meter.spent());
```
hoani: From my understanding, we want to separate contract logic from gas handler logic.
The reason is so that anyone who implements a gas handler can choose how to pay for gas, but we don't want that user to determine how a contract tracks gas spent.
If we put `finish` in the `GasHandler`, then the user still decides how the contract tracks gas.

In my opinion, to follow these rules, we should insert:
`GasSpent::mutate(|block_gas_spent| *block_gas_spent += gas_meter.spent());`
On line 592 (https://github.com/plugblockchain/plug-blockchain/blob/613e545f8ac6f07abdf8eacd7457c18fece6f0af/frame/contracts/src/lib.rs#L592)
and line 760 (https://github.com/plugblockchain/plug-blockchain/blob/613e545f8ac6f07abdf8eacd7457c18fece6f0af/frame/contracts/src/lib.rs#L760)
of the Contracts `lib.rs` so that full control is under the contract for how it tracks gas spent

hoani (changes requested):
Please see requested changes, happy to discuss design decisions =)

frame/contracts/src/gas.rs
jordy25519: We could do some kind of decorator but this would be so much simpler XD

aliXsed: The logic that updates the gas spent for the block is still relevant to the gas handler. The fact is a gas meter has two data. The gas spent that should be used to update the gas spent for the block and the gas left which is used in emptying unused gas. It seemed relevant to me to bundle these two into a function that would be used before finishing a contract as we would never want for one of these two to happen without the other one. 
On the other hand, the real aim was, we wanted to make sure if "empty_unused_gas" is overridden, the gas spent for the block would still keep being updated without exposing any new gas specific function of Plug to other projects like cennznet.

jordy25519 (approved):
TY LGTM

hoani (approved):
LGTM 👍

cowboy-bebug (approved):
LGTM !

PR #75, opened 2020-03-09
Attestation Module Audit Updates (BT-161)

hoani:
The following changes were made:

- Add fixed fee weights to `set_claim` and `remove_claim`
- Remove `set_self_claim` - there is a unit test which shows that a user can perform a self-claim using `set_claim` instead
- `ClaimSet` event has been replaced with `ClaimCreated` and `ClaimUpdated` to distinguish between the possible outcomes of `set_claim`
- `remove_claim` bug which removes an issuer who may have multiple claims on a holder has been resolved
- Will return `Error::TopicNotRegistered` when attempting to delete a non-existent topic
- Add doc comments
- Add 14 regression tests 

Concerns: 
- No regression tests added to cover fixed fee weights

### DIFF
--- a/prml/attestation/src/lib.rs
+++ b/prml/attestation/src/lib.rs
@@ -110,7 +110,7 @@ decl_storage! {
 		Topics get(fn get_topics):
 			map hasher(blake2_256) (T::AccountId, T::AccountId) => Vec<AttestationTopic>;
 		/// A map of (HolderId, IssuerId, AttestationTopic) => AttestationValue
-		Values get(fn get_values):
+		Values get(fn get_value):
 			map hasher(blake2_256) (T::AccountId, T::AccountId, AttestationTopic) => AttestationValue;
 	}
 }
@@ -145,27 +145,32 @@ impl<T: Trait> Module<T> {
 #[cfg(test)]
 mod tests {
 	use super::*;
-	use crate::mock::{ExtBuilder, Attestation, Test, Origin};
-	use sp_runtime::traits::Zero;
-	use frame_support::assert_noop;
+	use crate::mock::{ExtBuilder, Attestation, Origin};
 
 	#[test]
-	fn initialize_holder_has_no_issuers() {
+	fn initialize_holder_has_no_claims() {
 		let holder = 0xbaa;
 		ExtBuilder::build().execute_with(|| {
+			// Note: without any valid issuers, there is no valid input for
+			// get_topics or get_values
 			assert_eq!(Attestation::get_issuers(holder), []);
 		})
 	}
 
 	#[test]
-	fn adding_claim_mutates_issuers() {
+	fn adding_claim_to_storage() {
 		let issuer = 0xf00;
 		let holder = 0xbaa;
 		let topic = AttestationTopic::from(0xf00d);
 		let value = AttestationValue::from(0xb33f);
 		ExtBuilder::build().execute_with(|| {
-			Attestation::set_claim(Origin::signed(issuer), holder, topic, value);
+			let result = Attestation::set_claim(Origin::signed(issuer), holder, topic, value);
+
+			assert_eq!(result, Ok(()));
+
 			assert_eq!(Attestation::get_issuers(holder), [issuer]);
+			assert_eq!(Attestation::get_topics((holder, issuer)), [topic]);
+			assert_eq!(Attestation::get_value((holder, issuer, topic)), value);
 		})
 	}
 
@@ -178,9 +183,16 @@ mod tests {
 		let topic_loot = AttestationTopic::from(0x1007);
 		let value_loot = AttestationValue::from(0x901d);
 		ExtBuilder::build().execute_with(|| {
-			Attestation::set_claim(Origin::signed(issuer), holder, topic_food, value_food);
-			Attestation::set_claim(Origin::signed(issuer), holder, topic_loot, value_loot);
+			let result_food = Attestation::set_claim(Origin::signed(issuer), holder, topic_food, value_food);
+			let result_loot = Attestation::set_claim(Origin::signed(issuer), holder, topic_loot, value_loot);
+
+			assert_eq!(result_food, Ok(()));
+			assert_eq!(result_loot, Ok(()));
+
 			assert_eq!(Attestation::get_issuers(holder), [issuer]);
+			assert_eq!(Attestation::get_topics((holder, issuer)), [topic_food, topic_loot]);
+			assert_eq!(Attestation::get_value((holder, issuer, topic_food)), value_food);
+			assert_eq!(Attestation::get_value((holder, issuer, topic_loot)), value_loot);
 		})
 	}
 }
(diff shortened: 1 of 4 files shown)


prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
+        ///  holder: {
+        ///    issuer: {
+        ///      topic: <value>
+        ///    }
+        ///  }
+        /// }
+        ///
+
+        /// A map of HolderId => Vec<IssuerId>
+        Issuers get(fn get_issuers):
```
jordy25519: Rust convention is to not prefix getter methods with `get_`

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
-			Self::deposit_event(RawEvent::ClaimRemoved(holder, issuer, topic));
+            <Topics<T>>::mutate((holder.clone(), issuer.clone()),|topics| topics.retain(|vec_topic| *vec_topic != topic));
 
-			Ok(())
-		}
-	}
+            <Issuers<T>>::mutate(&holder, |issuers| {
+                    issuers.retain(|vec_issuer| {
+                        *vec_issuer != issuer.clone() ||
+                        Self::get_topics((holder.clone(), issuer.clone())).len() != 0
```
jordy25519: I don't think this check needs to run more than once for the `issuer` (can be done before the `retain` closure)
should read "if there are no more topics from this issuer, remove issuer, otherwise nothing"

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
+    /// Sets a claim about a `holder` from an `issuer`
+    /// If the claim `topic` already exists, then the claim `value` is updated,
+    /// Otherwise, a new claim is created for the `holder` by the `issuer`
+    fn create_or_update_claim(
+        holder: T::AccountId,
+        issuer: T::AccountId,
+        topic: AttestationTopic,
+        value: AttestationValue,
+    ) {
+        let is_update: bool = Self::get_topics((holder.clone(), issuer.clone())).contains(&topic);
```
jordy25519: Can this be declared below where it is used or maybe you meant to use it in `Topics<T>::mutate`?

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
+        let topic = AttestationTopic::from(0xf00d);
+        ExtBuilder::build().execute_with(|| {
+            assert_noop!(
+                Attestation::remove_claim(Origin::signed(issuer), holder, topic),
+                Error::<Test>::TopicNotRegistered
+            );
+        })
+    }
+
+    #[test]
```
jordy25519, line 459: TIOLI: could add the event checks in the previous unit tests

jordy25519 (commented):
TY this is a big improvement.
Some minor suggestions but otherwise LGTM

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
+    fn adding_existing_claim_overwrites_claim() {
+        let issuer = 0xf00;
+        let holder = 0xbaa;
+        let topic = AttestationTopic::from(0xf00d);
+        let value_old = AttestationValue::from(0xb33f);
+        let value_new = AttestationValue::from(0xcabba93);
+        ExtBuilder::build().execute_with(|| {
+            let result_old =
+                Attestation::set_claim(Origin::signed(issuer), holder, topic, value_old);
```
cowboy-bebug, line 232: [Optional] Do we want to assert whether `value_old` is set in storage after this call (just to show that the values do get replaced inside the storage):
```rust
assert_eq!(Attestation::get_value((holder, issuer, topic)), value_old);
```

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
+                Attestation::get_value((holder, issuer, topic_loot)),
+                value_loot
+            );
+        })
+    }
+
+    #[test]
+    fn adding_claims_from_different_issuers() {
+        let issuer_foo = 0xf00;
+        let issuer_boa = 0xb0a;
```
cowboy-bebug, line 275: 😄👍

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
+                topic,
+                value
+            ));
+
+            let expected_event =
+                TestEvent::attestation(RawEvent::ClaimCreated(holder, issuer, topic, value));
+            // Assert
+            assert!(System::events()
+                .iter()
+                .any(|record| record.event == expected_event));
```
cowboy-bebug, line 478: 👍

cowboy-bebug (approved):
LVGTM!

prml/attestation/src/lib.rs
hoani: Done, thanks for mentioning!

prml/attestation/src/lib.rs
hoani: Great catch, I have moved the check outside

prml/attestation/src/lib.rs
hoani: The idea here was to check whether we wanted to send an update event or a created event. You raise a good point though because a similar check is used in the topics closure. I will rearrange this a little bit to make the most of the checks we are doing

prml/attestation/src/lib.rs
hoani: Leaving it - I follow Kent Beck's testing pattern for one feature per test

prml/attestation/src/lib.rs
hoani: Nice idea, it makes the intent of the test very clear - Done

prml/attestation/src/lib.rs
hoani: ^ I am most proud of:
`topic = 0xf00d, values = b33f, 9047, cabba93`
😆

cowboy-bebug (approved):
The changes look good as well!

prml/attestation/src/lib.rs
```diff
@@ -35,102 +35,515 @@
...
-			<Values<T>>::remove((holder.clone(), issuer.clone(), topic));
+            <Values<T>>::remove((holder.clone(), issuer.clone(), topic));
 
-			Self::deposit_event(RawEvent::ClaimRemoved(holder, issuer, topic));
+            <Topics<T>>::mutate((holder.clone(), issuer.clone()),|topics| topics.retain(|vec_topic| *vec_topic != topic));
 
-			Ok(())
-		}
-	}
+            let remove_issuer = <Topics<T>>::get((holder.clone(), issuer.clone())).len() == 0;
```
jordy25519: zen mode ☯️  `<Topics<T>>::get((holder.clone(), issuer.clone())).len().is_zero()`

jordy25519 (approved):
0x13374ax

prml/attestation/src/lib.rs
hoani: ☯️ 
image

mzxyz (commented):
Awesome work, LGTM

PR #77, opened 2020-03-30
Chore/162 improve imbalance

mzxyz:
This PR try to improve the Positive/NegativeImbalance so they are fully compatible with our Generic Asset module

## Changes
* In `merge`, `subsume` and `offset` functions for both `NegativeImbalance` and `NegativeImbalance`:
    - imbalances without asset ID will take on the asset ID of input parameter.
    - These operation will not happen if asset_ids do not match

## Tests
- If imBalance is initialised by `zero` function, asset_id should be updated when calling `merge`, `subsume` or `offset` function.
- Should panic under debug env if asset_id is not consistent when calling `merge`, `subsume` or `offset` function.
- `total_issuance` should be updated after imBalance dropped.

### DIFF
--- a/frame/generic-asset/src/lib.rs
+++ b/frame/generic-asset/src/lib.rs
@@ -917,6 +917,24 @@ mod imbalances {
 	#[cfg_attr(test, derive(PartialEq, Debug))]
 	pub struct PositiveImbalance<T: Subtrait>(T::Balance, T::AssetId);
 	impl<T: Subtrait> PositiveImbalance<T> {
+		/// This is a helper function that checks the consistency of asset ID for operations on Imbalances
+		///
+		/// If the imblance a has no asset_id configured ( asset_id == 0 ), a takes on b's asset_id
+		///
+		/// If a and b do not have matching asset_ids, debug_assert and return false.
+		/// Otherwise return true
+		fn check_asset_id_consistency(&mut self, other: &Self) -> bool {
+			let mut result = true;
+			if self.1 == Zero::zero() {
+				self.1 = other.1;
+			}
+
+			if self.1 != other.1 {
+				debug_assert!(false, "Asset ID do not match!");
+				result = false;
+			}
+			result
+		}
 		pub fn new(amount: T::Balance, asset_id: T::AssetId) -> Self {
 			PositiveImbalance(amount, asset_id)
 		}
@@ -928,6 +946,25 @@ mod imbalances {
 	#[cfg_attr(test, derive(PartialEq, Debug))]
 	pub struct NegativeImbalance<T: Subtrait>(T::Balance, T::AssetId);
 	impl<T: Subtrait> NegativeImbalance<T> {
+		/// This is a helper function that checks the consistency of asset ID for operations on Imbalances
+		///
+		/// If the imblance a has no asset_id configured ( asset_id == 0 ), a takes on b's asset_id
+		///
+		/// If a and b do not have matching asset_ids, debug_assert and return false.
+		/// Otherwise return true
+		fn check_asset_id_consistency(&mut self, other: &Self) -> bool {
+			let mut result = true;
+			if self.1 == Zero::zero() {
+				self.1 = other.1;
+			}
+
+			if self.1 != other.1 {
+				debug_assert!(false, "Asset ID do not match!");
+				result = false;
+			}
+			result
+		}
+
 		pub fn new(amount: T::Balance, asset_id: T::AssetId) -> Self {
 			NegativeImbalance(amount, asset_id)
 		}
@@ -963,23 +1000,34 @@ mod imbalances {
 			(Self::new(first, asset_id), Self::new(second, asset_id))
 		}
 		fn merge(mut self, other: Self) -> Self {
-			self.0 = self.0.saturating_add(other.0);
-			mem::forget(other);
-
+			// Only merge when asset_id match
+			if self.check_asset_id_consistency(&other) {
+				self.0 = self.0.saturating_add(other.0);
+				mem::forget(other);
+			}
 			self
 		}
 		fn subsume(&mut self, other: Self) {
-			self.0 = self.0.saturating_add(other.0);
-			mem::forget(other);
+			// Only subsume when asset_id match
+			if self.check_asset_id_consistency(&other) {
+				self.0 = self.0.saturating_add(other.0);
+				mem::forget(other);
+			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let (a, b, asset_id) = (self.0, other.0, self.1);
-			mem::forget((self, other));
-
-			if a >= b {
-				Ok(Self::new(a - b, asset_id))
+			let asset_id = if self.1 == Zero::zero() { other.1 } else { self.1 };
+			if asset_id != other.1 {
+				debug_assert!(false, "Asset ID do not match!");
+				Ok(self)
 			} else {
-				Err(NegativeImbalance::new(b - a, asset_id))
+				let (a, b) = (self.0, other.0);
+				mem::forget((self, other));
+
+				if a >= b {
+					Ok(Self::new(a - b, asset_id))
+				} else {
+					Err(NegativeImbalance::new(b - a, asset_id))
+				}
 			}
 		}
 		fn peek(&self) -> T::Balance {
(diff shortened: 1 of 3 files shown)


frame/generic-asset/src/lib.rs
```diff
@@ -911,6 +911,34 @@ mod imbalances {
 	};
 	use sp_std::mem;
 
+	/// Trait used to avoid duplicate code that is implemented for both the Positive and Negative
```
hoani: [suggestion] I would remove the first two lines of these comments - the third line describes the purpose of this trait really nicely 😄

frame/generic-asset/src/lib.rs
```diff
@@ -911,6 +911,34 @@ mod imbalances {
...
+		///
+		/// If a and b do not have matching asset_ids, debug_assert and return false.
+		/// Otherwise return true
+		fn match_asset_id(&mut self, other: &Self) -> bool {
+			let mut result = true;
+			if self.asset_id() == Zero::zero() {
+				self.set_asset_id(other.asset_id().clone());
+			}
+
+			if self.asset_id() != other.asset_id() {
```
hoani: [suggestion] tempted to just go `return self.asset_id() == other.asset_id()`, it depends on how much we want the `debug_asset!`

frame/generic-asset/src/lib.rs
```diff
@@ -963,23 +1015,34 @@ mod imbalances {
...
+			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let (a, b, asset_id) = (self.0, other.0, self.1);
-			mem::forget((self, other));
-
-			if a >= b {
-				Ok(Self::new(a - b, asset_id))
+			let asset_id = if self.1 == Zero::zero() { other.1 } else { self.1 };
+			if asset_id != other.1 {
```
hoani, line 1026: Shouldn't we be using `self.match_asset_id(&other)` here?
Then we can just use `self.1` instead of `asset_id` below

frame/generic-asset/src/lib.rs
```diff
@@ -1015,23 +1078,35 @@ mod imbalances {
...
```
hoani, line 1090: ^ Same comment as above

frame/generic-asset/src/tests.rs
```diff
@@ -1245,3 +1246,172 @@ fn can_set_asset_owner_permissions_in_genesis() {
...
+			let other = NegativeImbalanceOf::new(100, asset_id);
+			assert_eq!(negative_im.asset_id(), 0);
+			assert_eq!(negative_im.peek(), 0);
+			assert_eq!(other.asset_id(), asset_id);
+			// merge
+			let merged_im = negative_im.merge(other);
+			assert_eq!(merged_im.asset_id(), asset_id);
+			assert_eq!(merged_im.peek(), 100);
+			// subsume
+			let mut negative_im = NegativeImbalanceOf::zero();
```
hoani: [nitpick] can remove `mut` here

frame/generic-asset/src/tests.rs
```diff
@@ -1245,3 +1246,172 @@ fn can_set_asset_owner_permissions_in_genesis() {
...
+			let opposite_im = NegativeImbalanceOf::new(50, asset_id);
+			let offset_im = negative_im.offset(opposite_im).unwrap();
+			assert_eq!(offset_im.asset_id(), asset_id);
+			assert_eq!(offset_im.peek(), 50);
+	});
+}
+
+#[test]
+#[should_panic(expected = "Asset ID do not match!")]
+fn negative_imbalance_merge_with_imcompatible_asset_id_should_panic() {
```
hoani: [suggestion] name these to `..._should_error` we don't want to panic ;)

frame/generic-asset/src/tests.rs
```diff
@@ -1245,3 +1246,172 @@ fn can_set_asset_owner_permissions_in_genesis() {
...
+	ExtBuilder::default().build().execute_with(|| {
+		// create new positive imbalance
+		let negative_im = NegativeImbalanceOf::new(100, 1);
+		let other = NegativeImbalanceOf::new(50, 2);
+		// merge
+		let _ = negative_im.merge(other);
```
hoani: did we miss an `assert_err()` here?

frame/generic-asset/src/tests.rs
```diff
@@ -1245,3 +1246,172 @@ fn can_set_asset_owner_permissions_in_genesis() {
...
+			assert_eq!(GenericAsset::total_issuance(&asset_id), balance);
+			// generate empty positive imbalance
+			let positive_im = PositiveImbalanceOf::zero();
+			let other = PositiveImbalanceOf::new(100, asset_id);
+			// merge
+			let merged_im = positive_im.merge(other);
+			// explitically drop `imbalance` so issuance is managed
+			drop(merged_im);
+			assert_eq!(GenericAsset::total_issuance(&asset_id), balance + 100);
+	});
```
hoani, line 1490: Very nice 👍

frame/generic-asset/src/tests.rs
```diff
@@ -1245,3 +1246,172 @@ fn can_set_asset_owner_permissions_in_genesis() {
 			assert_eq!(expected, actual);
 	});
 }
+
+#[test]
+fn zero_asset_id_should_updated_after_negative_imbalance_operations() {
```
hoani, line 1250: We are currently covering the cases:
* `asset_id(A) = 0, asset_id(B) = 16000`
I think it would be good to add the cases 
* `asset_id(A) = 16000, asset_id(B) = 16000`
This could be done by merging, subsuming and offsetting twice in this test

hoani (changes requested):
Great work so far @HaoZhaiX @royyang-centrality 
I have added some comments for you to consider, overall the design and test coverage is great

frame/generic-asset/src/lib.rs
roy-powerfinance: Got it

frame/generic-asset/src/lib.rs
roy-powerfinance: I think it's important to assert because operating on imbalances with different asset ID definitely indicate bugs I the code and we should catch it early

frame/generic-asset/src/lib.rs
roy-powerfinance: We can't - the "self" is not passed in as a mutable reference so we can't change the asset id.
In addition there's no point because the code  "forget" both the "self" and "other" and construct a "new" imbalance to be returned.

frame/generic-asset/src/lib.rs
hoani: Fair enough, I'm happy either way 😄

frame/generic-asset/src/lib.rs
hoani: Good point 👍

frame/generic-asset/src/tests.rs
hoani: Resolving this - turns out we might be ok with panicking

roy-powerfinance: currently the wrong asset_id will cause a debug_panic - we will panic to indicate a important bug, but only in debug environment

mzxyz: Good suggestion, just comparing to other test code which using `should_panic`, will change it to `should_failed`

frame/generic-asset/src/tests.rs
mzxyz: Good point, add check for imbalance operations with same asset id

hoani (approved):
Nice updates, LGTM

frame/generic-asset/src/lib.rs
```diff
@@ -911,6 +911,32 @@ mod imbalances {
...
+
+		/// This is a helper function that checks the consistency of asset ID for operations on Imbalances
+		///
+		/// If the imbalance a has no asset_id configured ( asset_id == 0 ), a takes on b's asset_id
+		///
+		/// If a and b do not have matching asset_ids, debug_assert and return false.
+		/// Otherwise return true
+		fn match_asset_id(&mut self, other: &Self) -> bool {
+			let mut result = true;
+			if self.get_asset_id() == Zero::zero() {
```
cowboy-bebug, line 927: I think it's possible to create an asset with `asset_id = 0`, which will be replaced by `other`. It might be worthwhile to disallow creating an asset with id = 0.

cowboy-bebug (approved):
LGTM! I have just one comment which could be addressed in another PR

frame/generic-asset/src/lib.rs
mzxyz: Yes, we could let `Next_Asset_id` not start from `zero`

hoani: We will also need to make sure `asset_id` can not be 0 in `fn create_reserved`

roy-powerfinance: jira created here: 
https://centralitydev.atlassian.net/browse/BT-184

PR #78, opened 2020-04-08
Made some improvements according to the "General" section of the Generic

roy-powerfinance:
General section:
* Made asset id of 0 the "null_asset_id"
* Improved some error names
* Check for target account overflow in make_transfer

Imbalance section
* Asset Id for imbalances cannot be changed once it is set
* Imbalances now have named fields
* Removed the InherentAssetIdProvider

### DIFF
--- a/frame/generic-asset/src/lib.rs
+++ b/frame/generic-asset/src/lib.rs
@@ -191,7 +191,6 @@ pub trait Trait: frame_system::Trait {
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
 }
 
-
 pub trait Subtrait: frame_system::Trait {
 	type Balance: Parameter
 		+ Member
--- a/frame/generic-asset/src/lib.rs
+++ b/frame/generic-asset/src/lib.rs
@@ -602,7 +602,7 @@ impl<T: Trait> Module<T> {
 		options: AssetOptions<T::Balance, T::AccountId>,
 	) -> DispatchResult {
 		let asset_id = if let Some(asset_id) = asset_id {
-			ensure!(asset_id != Zero::zero(),  Error::<T>::AssetIdUnavailable);
+			ensure!(!asset_id.is_zero(),  Error::<T>::AssetIdUnavailable);
 			ensure!(!<TotalIssuance<T>>::contains_key(&asset_id), Error::<T>::AssetIdAlreadyTaken);
 			ensure!(asset_id < Self::next_asset_id(), Error::<T>::AssetIdUnavailable);
 			asset_id
@@ -931,14 +931,15 @@ mod imbalances {
 		/// Otherwise return true
 		fn match_asset_id(&mut self, other: &Self) -> bool {
 			let mut result = true;
-			match self.get_asset_id() {
-				None => self.set_asset_id(other.get_asset_id()),
-				Some(asset_id) => {
-					if asset_id != other.get_asset_id().unwrap_or(Zero::zero()) {
+			match (self.get_asset_id(), other.get_asset_id()) {
+				(None, Some(asset_id)) => self.set_asset_id(other.get_asset_id()),
+				(Some(this_asset), Some(other_asset)) => {
+					if this_asset != other_asset {
 						debug_assert!(false, "Asset ID do not match!");
 						result = false;
 					}
 				},
+				_ => {},
 			}
 			result
 		}
@@ -1030,7 +1031,7 @@ mod imbalances {
 			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let asset_id = if self.1 == None { other.1 } else { self.1 };
+			let asset_id = if self.1.is_none() { other.1 } else { self.1 };
 			if asset_id != other.1 {
 				debug_assert!(false, "Asset ID do not match!");
 				Ok(self)
@@ -1094,7 +1095,7 @@ mod imbalances {
 			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let asset_id = if self.1 == None { other.1 } else { self.1 };
+			let asset_id = if self.1.is_none() { other.1 } else { self.1 };
 			if asset_id != other.1 {
 				debug_assert!(false, "Asset ID do not match!");
 				Ok(self)
@@ -1117,9 +1118,8 @@ mod imbalances {
 	impl<T: Subtrait> Drop for PositiveImbalance<T> {
 		/// Basic drop handler will just square up the total issuance.
 		fn drop(&mut self) {
-			match self.1 {
-				Some(asset_id) => <super::TotalIssuance<super::ElevatedTrait<T>>>::mutate(asset_id, |v| *v = v.saturating_add(self.0)),
-				None => (),
+			if let Some(asset_id) = self.1 {
+				<super::TotalIssuance<super::ElevatedTrait<T>>>::mutate(asset_id, |v| *v = v.saturating_add(self.0));
 			}
 		}
 	}
@@ -1127,9 +1127,8 @@ mod imbalances {
 	impl<T: Subtrait> Drop for NegativeImbalance<T> {
 		/// Basic drop handler will just square up the total issuance.
 		fn drop(&mut self) {
-			match self.1 {
-				Some(asset_id) => <super::TotalIssuance<super::ElevatedTrait<T>>>::mutate(&asset_id, |v| *v = v.saturating_sub(self.0)),
-				None => (),
+			if let Some(asset_id) = self.1 {
+				<super::TotalIssuance<super::ElevatedTrait<T>>>::mutate(&asset_id, |v| *v = v.saturating_sub(self.0));
 			}
 		}
 	}
--- a/frame/generic-asset/src/lib.rs
+++ b/frame/generic-asset/src/lib.rs
@@ -921,9 +921,8 @@ mod imbalances {
 
 	/// Provide access to asset ID within imbalance structs
 	pub trait ImbalanceWithAssetId<T: Subtrait>{
-		fn null_asset_id() -> T::AssetId{ Zero::zero() }
-		fn get_asset_id(&self) -> T::AssetId;
-		fn set_asset_id(&mut self, asset_id : T::AssetId);
+		fn get_asset_id(&self) -> Option<T::AssetId>;
+		fn set_asset_id(&mut self, asset_id : Option<T::AssetId>);
 
 		/// This is a helper function that checks the consistency of asset ID for operations on Imbalances
 		///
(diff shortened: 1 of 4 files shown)


frame/generic-asset/src/lib.rs
```diff
@@ -191,6 +191,7 @@ pub trait Trait: frame_system::Trait {
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Trait>::Event>;
 }
 
+
```
jordy25519: Whitespace

frame/generic-asset/src/lib.rs
```diff
@@ -913,6 +921,7 @@ mod imbalances {
 
 	/// Provide access to asset ID within imbalance structs
 	pub trait ImbalanceWithAssetId<T: Subtrait>{
+		fn null_asset_id() -> T::AssetId{ Zero::zero() }
```
jordy25519: Interesting, should our GA imbalance actually have it's `asset_id` type as`Option<AssetId>`?
Rust doesn't have null but Option, and this feels like it's semantically what we want to do here

frame/generic-asset/src/lib.rs
```diff
@@ -989,7 +998,7 @@ mod imbalances {
 		fn zero() -> Self {
 			// We lose asset /currency ID information here
 			// It should be fine unless we are trying to reconstruct that information from a zeroed Imbalance
-			Self::new(Zero::zero(), Zero::zero())
+			Self::new(Zero::zero(), Self::null_asset_id())
```
jordy25519: this would be `Self::new(Zero::zero(), None)` if doing the Optional thing

jordy25519 (commented):
Error changes look good.
I feel the 'null' asset_id could be designed more rustically

frame/generic-asset/src/lib.rs
roy-powerfinance: yeah that sounds good!

mzxyz: Change may like:
```
pub struct PositiveImbalance<T: Subtrait>(T::Balance, Option(T::AssetId));
```
But not sure it's good idea to change zero fun:
```
fn zero() -> Self {
  Self::new(Zero::zero(), None)
}
```

frame/generic-asset/src/lib.rs
roy-powerfinance: Fixed

frame/generic-asset/src/lib.rs
roy-powerfinance: Changed the imbalance structs and interfaces to Option<AssetId>

frame/generic-asset/src/lib.rs
```diff
@@ -1086,7 +1094,7 @@ mod imbalances {
 			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let asset_id = if self.1 == Zero::zero() { other.1 } else { self.1 };
+			let asset_id = if self.1 == None { other.1 } else { self.1 };
```
jordy25519:
```suggestion
			let asset_id = if self.1.is_none() { other.1 } else { self.1 };
```

frame/generic-asset/src/lib.rs
```diff
@@ -1109,27 +1117,33 @@ mod imbalances {
 	impl<T: Subtrait> Drop for PositiveImbalance<T> {
 		/// Basic drop handler will just square up the total issuance.
 		fn drop(&mut self) {
-			<super::TotalIssuance<super::ElevatedTrait<T>>>::mutate(&self.1, |v| *v = v.saturating_add(self.0));
+			match self.1 {
+				Some(asset_id) => <super::TotalIssuance<super::ElevatedTrait<T>>>::mutate(asset_id, |v| *v = v.saturating_add(self.0)),
```
jordy25519: Could change these to `if let Some(asset_id) {` format

frame/generic-asset/src/lib.rs
```diff
@@ -1022,7 +1030,7 @@ mod imbalances {
 			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let asset_id = if self.1 == Zero::zero() { other.1 } else { self.1 };
+			let asset_id = if self.1 == None { other.1 } else { self.1 };
```
jordy25519:
```suggestion
			let asset_id = if self.1.is_none() { other.1 } else { self.1 };
```

frame/generic-asset/src/lib.rs
```diff
@@ -600,14 +602,15 @@ impl<T: Trait> Module<T> {
 		options: AssetOptions<T::Balance, T::AccountId>,
 	) -> DispatchResult {
 		let asset_id = if let Some(asset_id) = asset_id {
-			ensure!(!<TotalIssuance<T>>::contains_key(&asset_id), Error::<T>::IdAlreadyTaken);
-			ensure!(asset_id < Self::next_asset_id(), Error::<T>::IdUnavailable);
+			ensure!(asset_id != Zero::zero(),  Error::<T>::AssetIdUnavailable);
```
jordy25519:
```suggestion
			ensure!(!asset_id.is_zero(),  Error::<T>::AssetIdUnavailable);
```

frame/generic-asset/src/lib.rs
```diff
@@ -924,13 +931,14 @@ mod imbalances {
...
 		fn match_asset_id(&mut self, other: &Self) -> bool {
 			let mut result = true;
-			if self.get_asset_id() == Zero::zero() {
-				self.set_asset_id(other.get_asset_id().clone());
-			}
-
-			if self.get_asset_id() != other.get_asset_id() {
-				debug_assert!(false, "Asset ID do not match!");
-				result = false;
+			match self.get_asset_id() {
```
jordy25519: Can match over the combos here to make it really clear what we're doing
```
match (self.1, other.get_asset_id()) {
    None, Some => // replace self.1 ok
    Some, None => // do nothing? not sure whats best here
    Some, Some => // must be the same for any valid op. debug_assert!
    None, None =>  // nothing to do here.
}
```

jordy25519 (commented):
🚀 this looks really good. Some small style things and then I think we're good to go

frame/generic-asset/src/lib.rs
roy-powerfinance: Yep

frame/generic-asset/src/lib.rs
roy-powerfinance: this is cool. done

frame/generic-asset/src/lib.rs
roy-powerfinance: fixed

frame/generic-asset/src/lib.rs
roy-powerfinance: fixed

frame/generic-asset/src/lib.rs
roy-powerfinance: ha this is really cool. fixed!

frame/generic-asset/src/lib.rs
```diff
@@ -1022,7 +1031,7 @@ mod imbalances {
 			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let asset_id = if self.1 == Zero::zero() { other.1 } else { self.1 };
+			let asset_id = if self.1.is_none() { other.1 } else { self.1 };
```
mzxyz: using `unwrap_or_else` instead of `if else`
```
self.1.unwrap_or_else(|| other.1)
```

frame/generic-asset/src/lib.rs
```diff
@@ -924,13 +931,15 @@ mod imbalances {
...
 		fn match_asset_id(&mut self, other: &Self) -> bool {
 			let mut result = true;
-			if self.get_asset_id() == Zero::zero() {
-				self.set_asset_id(other.get_asset_id().clone());
-			}
-
-			if self.get_asset_id() != other.get_asset_id() {
-				debug_assert!(false, "Asset ID do not match!");
-				result = false;
+			match (self.get_asset_id(), other.get_asset_id()) {
```
mzxyz: [optional] change to `match (self.1, other.1)`

roy-powerfinance: this can't be done - the trait doesn't know about class structure - so it has to rely on the trait function to get its asset_id

frame/generic-asset/src/lib.rs
roy-powerfinance: the asset _id is of the type option<assetId>, because it can be None. So if we use unwrap, we have to re-construct the "Option", or use Zero::zero() as the placeholder for None.
i think it's better to keep asset_id here as the Option<assetId> format

mzxyz: That's make sense. In that case we can use `self.1.or(other.1)`

roy-powerfinance: yep good idea

frame/generic-asset/src/lib.rs
```diff
@@ -336,7 +336,7 @@ decl_error! {
 	/// Error for the generic-asset module.
 	pub enum Error for Module<T: Trait> {
 		/// No new assets id available.
-		NoIdAvailable,
+		NoAssetIdAvailable,
```
jordy25519: This should be `AssetIdExhausted`

frame/generic-asset/src/lib.rs
```diff
@@ -354,11 +354,13 @@ decl_error! {
 		/// Free balance got underflowed after burning.
 		FreeBurningUnderflow,
 		/// Asset id is already taken.
-		IdAlreadyTaken,
+		AssetIdAlreadyTaken,
```
jordy25519: `AssetIdExists`

frame/generic-asset/src/lib.rs
```diff
@@ -1052,46 +1073,45 @@ mod imbalances {
...
 			if self.match_asset_id(&other) {
-				self.0 = self.0.saturating_add(other.0);
+				self.amount = self.amount.saturating_add(other.amount);
 				mem::forget(other);
 			}
 		}
 		fn offset(self, other: Self::Opposite) -> result::Result<Self, Self::Opposite> {
-			let asset_id = if self.1 == Zero::zero() { other.1 } else { self.1 };
-			if asset_id != other.1 {
+			let asset_id = if self.asset_id.is_none() { other.asset_id } else { self.asset_id };
```
mzxyz: Same with the change: `self.asset_id.or(other.asset_id)`

mzxyz:
Awesome work!  There are 3 things need to be fixed otherwise LGTM:
- Fix the last comment
- Fix the failed test
- Rebase from develop

frame/generic-asset/src/lib.rs
jordy25519: should do the `.or` thing here too

frame/generic-asset/src/tests.rs
```diff
@@ -22,8 +22,10 @@
 
 use super::*;
 use crate::mock::{new_test_ext, ExtBuilder, GenericAsset, Origin, System, Test, TestEvent, PositiveImbalanceOf, NegativeImbalanceOf};
+use crate::imbalances::ImbalanceWithAssetId;
 use frame_support::{assert_noop, assert_ok, traits::Imbalance};
 
+
```
jordy25519, line 28: White space

jordy25519 (approved):
some last nitpicks but otherwise GJ 👍 
It's a big improvement!

mzxyz (approved):

PR #81, opened 2020-04-10
Add delegated gas fee test to contracts

hoani:
Add a simple delegated gas fee test to show that the issuer is charged for gas on contract calls

## Changes:

- Add test

## Concerns:

- None

### DIFF
--- a/frame/contracts/src/doughnut_integration.rs
+++ b/frame/contracts/src/doughnut_integration.rs
@@ -702,3 +702,32 @@ fn delegated_runtime_to_contract_call_returns_error_with_unverifiable_doughnut()
 		);
 	});
 }
+
+/// Simple test to ensure that gas charges are made to the issuer
+/// (who becomes the caller of a delegated origin)
+#[test]
+fn contract_call_charges_gas_to_issuer() {
+	let verifiable_doughnut = MockDoughnut::default()
+		.set_runtime_verifiable(true);
+	let delegated_origin = RawOrigin::from((Some(ALICE), Some(verifiable_doughnut.clone())));
+	let gas_limit = 200_000;
+	let initial_balance = 1_000_000;
+	let gas_cost = 970;
+
+	ExtBuilder::default().existential_deposit(50).build().execute_with(|| {
+		Balances::deposit_creating(&ALICE, initial_balance);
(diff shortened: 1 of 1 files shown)


mzxyz (approved):
Nice, LGTM

frame/contracts/src/doughnut_integration.rs
```diff
@@ -702,3 +702,32 @@ fn delegated_runtime_to_contract_call_returns_error_with_unverifiable_doughnut()
 		);
 	});
 }
+
+/// Simple test to ensure that gas charges are made to the issuer
+/// (who becomes the caller of a delegated origin)
+#[test]
+fn contract_call_charges_gas_to_issuer() {
+	let verifiable_doughnut = MockDoughnut::default()
```
jordy25519, line 710: Abit tricksy that we can't see the issuer from the doughnut setup step easily

frame/contracts/src/doughnut_integration.rs
```diff
@@ -702,3 +702,32 @@ fn delegated_runtime_to_contract_call_returns_error_with_unverifiable_doughnut()
...
+	let delegated_origin = RawOrigin::from((Some(ALICE), Some(verifiable_doughnut.clone())));
+	let gas_limit = 200_000;
+	let initial_balance = 1_000_000;
+	let gas_cost = 970;
+
+	ExtBuilder::default().existential_deposit(50).build().execute_with(|| {
+		Balances::deposit_creating(&ALICE, initial_balance);
+		// Call BOB contract, which attempts to instantiate and call the callee contract
+		assert_ok!(Contract::call(
+			delegated_origin.clone().into(),
```
jordy25519: not sure we need clone here

jordy25519 (approved):
TY, some minor things

cowboy-bebug (approved):
Nice

frame/contracts/src/doughnut_integration.rs
hoani: Agree, it's because the mock doughnut does not hold the issuer, so you just need to assume that the issuer has been delegated the caller in this test 😞

frame/contracts/src/doughnut_integration.rs
hoani: Nice, removed it!

PR #85, opened 2020-04-18
Add Dispatch Method Arguments

hoani:
Adds runtime method arguments to `verify_dispatch`.

Arguments are added as `Vec<(&str, &dyn Any)>`.

Where the first argument in the tuple is a string indicating the type of the argument, and the second argument stored the value generically using the `Any` trait. The value can be extracted using `.downcast_ref::<T>()` in the dispatch verifier

For example a call to balances transfer gives:
```rust
[
  ("T::AccountId",  value as &dyn Any),
  ("T::Balance", value as &dyn Any),
]
```

It is the responsibility of the `DelegatedDispatchVerifier` to determine the best way to represent the data based on the application.

## Changes

* Add a 4th argument, `args: Vec<&str, &dyn Any>` to `DelegatedDispatchVerifier::verify_dispatch(...)`
* Leverage SCALE codec to encode the values
* Add unit test showing how type inference can be achieved

## Concerns

* Some runtime modules use `<T::Lookup as StaticLookup>::Source` (for example, `Balances::transfer`), in these cases it will be difficult to infer the type from the string.
  * I don't know if there is a reasonable way around this - `<T::Lookup as StaticLookup>::Source` is too ambiguous.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -31,7 +31,7 @@ commands:
     steps:
       - restore_cache:
           name: Restore sccache
-          key: sccache-{{ arch }}-{{ .Environment.CIRCLE_JOB }}
+          key: sccache-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-0
   save-cache:
     steps:
       - save_cache:
@@ -40,7 +40,7 @@ commands:
           # Of course, restore_cache will not find this exact key,
           # but it will fall back to the closest key (aka the most recent).
           # See https://discuss.circleci.com/t/add-mechanism-to-update-existing-cache-key/9014/13
-          key: sccache-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-{{ epoch }}
+          key: sccache-{{ arch }}-{{ .Environment.CIRCLE_JOB }}-0-{{ epoch }}
           paths:
             - "~/.cache/sccache"
   save-target-cache:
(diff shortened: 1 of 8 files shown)


.circleci/config.yml
```diff
@@ -13,7 +13,7 @@ commands:
             source ~/.cargo/env
             rustup install $RUST_VERSION
             rustup default stable
-            rustup install nightly
+            rustup install nightly-2020-04-17
             rustup target add wasm32-unknown-unknown --toolchain=nightly
```
jordy25519: I think build failure might resolve by changing toolchain name here:
`rustup target add wasm32-unknown-unknown --toolchain=nightly-2020-04-17`

hoani: Thanks 😄

roy-powerfinance (approved):
My only question is : According to Rust doc, Any trait had a "fn type_id(&self) -> TypeId"
what was the reason to use a string type identifier instead of using the trait?
Also it looks to me like type_id wasn't implemented, is the trait derived and have a default implementation?

Otherwise the PR looks good :D

frame/support/src/dispatch.rs
```diff
@@ -1407,16 +1407,25 @@ macro_rules! decl_module {
 						// Trait imports for doughnut dispatch verification
 						use $crate::additional_traits::MaybeDoughnutRef;
 						use $crate::dispatch::DelegatedDispatchVerifier;
+						use $crate::sp_std::{prelude::Vec, any::Any};
+						use $crate::codec::Encode;
 						// Check whether `origin` is acting with delegated authority (i.e. doughnut attached).
 						if let Some(doughnut) = &$from.doughnut() {
+							// Write arguments as a vector of strings
+							// Leverages off enforcement of Debug formatting on all extrinsic parameter types.
+							let mut arguments = Vec::<(&str, &dyn Any)>::default();
```
jordy25519: with some macro magic may be able to do `Vec::with_capacity` here

jordy25519 (approved):
lgtm!! would recommend undoing the CI changes now that we know it passes

frame/support/src/dispatch.rs
hoani: Done :)

PR #89, opened 2020-06-02
Scale down the stored balances

### LINKED ISSUE #88: Add a feature to Generic Asset to allow for correcting the scales
Generic Asset should allow correcting the previously-stored balances for a particular asset id. This can only happen on a runtime upgrade and based on the genesis config.

aliXsed:
Based on the individual scale factors coming from the genesis configuration for some asset ids.
Closes #88.

Tested and ready to be merged.

### DIFF
--- a/frame/generic-asset/src/lib.rs
+++ b/frame/generic-asset/src/lib.rs
@@ -528,7 +528,7 @@ decl_storage! {
 		pub SpendingAssetId get(fn spending_asset_id) config(): T::AssetId;
 
 		/// Temporary code for correcting the scale of balances for some assets
-		pub RuntimeUpgradeScalers get(fn runtime_upgrade_scalers) config():
+		pub RuntimeUpgradeScalers get(fn runtime_upgrade_scalers):
 			map hasher(twox_64_concat) T::AssetId => T::Balance;
 	}
 	add_extra_genesis {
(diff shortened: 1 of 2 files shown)


aliXsed:
@holygits I have fixed the formatting of test.rs unintentionally as well. So if it's not making it too hard for you to review the change let's keep it, otherwise please let me know and I will revert formatting changes.

jordy25519:
Please revert, best to stay consistent and use tabbed indentation.

The burning process lgtm.

We'd still need some way to set the values of the 'Scalers' e.g. new extrinsic call.

since we're scaling by the same factor (and it's one off), it may be simpler to use a runtime constant instead like:
https://github.com/cennznet/cennznet/blob/89d2688b048bb64aa5ec4f9e70bb59839dd9b352/runtime/src/lib.rs#L292

aliXsed:
> Please revert, best to stay consistent and use tabbed indentation.
> 
> The burning process lgtm.
> 
> We'd still need some way to set the values of the 'Scalers' e.g. new extrinsic call.
> 
> since we're scaling by the same factor (and it's one off), it may be simpler to use a runtime constant instead like:
> https://github.com/cennznet/cennznet/blob/89d2688b048bb64aa5ec4f9e70bb59839dd9b352/runtime/src/lib.rs#L292

@holygits Setting the scalers would be an individual PR for cennznet. I have prepared the change and I'm now in the process of debugging it. However, that change wouldn't be compilable until we merge this change, release Plug and change Cennznet to use the latest Plug.

aliXsed:
@holygits regarding the original formatting of test.rs, the indentations in the file is already a mix of space and tabs. I guess without formatting Plug, it's hard to maintain formatting and the inconsistency would just get worse. I know we haven't formatted this project to be compatible with the upstream, but we cal always force for all cherry-picks to go through formatting first. I suggest we decide about this in our team meetings. 
Anyway, I have reverted the formatting changes for this PR.

aliXsed:
We've managed to contain the whole logic for this issue in a change for cennznet: https://github.com/cennznet/cennznet/pull/229. So I'm going to close this one.

PR #91, opened 2020-06-04
Add a "rpc-disable-reserved-peers" feature flag to sc-rpc

hoani:
This PR allows us to optionally disable reserved peer client rpc calls.

It adds the feature flag `rpc-disable-reserved-peers` to the sc-rpc, which is not set by default.

When `rpc-disable-reserved-peers` is not set, then the rpc call behaves as normal:
```
hoani@Hoani-MBP plug-blockchain % curl -H "Content-Type: application/json" --data '{"id": 1,"jsonrpc": "2.0","method": "system_addReservedPeer","params": ["ws://hoani-node.node"]}' localhost:9933
{"jsonrpc":"2.0","error":{"code":2002,"message":"MultiaddrParse(InvalidMultiaddr)"},"id":1}
```

When `rpc-disable-reserved-peers` is set like this (in `bin/node/cli/Cargo.toml`):
```
sc-rpc = { version = "2.0.0-alpha.5", path = "../../../client/rpc", features=["rpc-disable-reserved-peers"] }
```

Then the RPC call is rejected:
```
hoani@Hoani-MBP plug-blockchain % curl -H "Content-Type: application/json" --data '{"id": 1,"jsonrpc": "2.0","method": "system_addReservedPeer","params": ["ws://hoani-node.node"]}' localhost:9933
{"jsonrpc":"2.0","error":{"code":2003,"message":"Reserved Peer Commands Disabled"},"id":1}
```

## Changes
* Add feature flag `rpc-disable-reserved-peers` to `sc-rpc`
* Add error type `Error::ReservedPeerCommandsDisabled` to `sc_rpc_api::system::error`
* Add feature check to disable the `removed_reserved_peer` and `add_reserved_peer` calls

## Concerns
* I would have preferred to completely disable the rpc calls based on feature flags, but given that the `#[rpc]` attributes were being processed before the `#[cfg_attr...]`, this wouldn't work.

### DIFF
--- a/client/rpc/Cargo.toml
+++ b/client/rpc/Cargo.toml
@@ -9,6 +9,7 @@ repository = "https://github.com/paritytech/substrate/"
 description = "Substrate Client RPC"
 
 [features]
+# Disables system_addReservedPeer and system_removeReservedPeer rpc methods
 rpc-disable-reserved-peers = []
 default = []
 
--- a/client/rpc/Cargo.toml
+++ b/client/rpc/Cargo.toml
@@ -9,9 +9,8 @@ repository = "https://github.com/paritytech/substrate/"
 description = "Substrate Client RPC"
 
 [features]
-rpc-config-reserved-peers = ["sc-rpc-api/rpc-config-reserved-peers"]
-
-default = ["rpc-config-reserved-peers"]
+rpc-disable-reserved-peers = []
+default = []
 
 [dependencies]
 sc-rpc-api = { version = "0.8.0-alpha.5", path = "../rpc-api" }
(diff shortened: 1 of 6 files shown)


client/rpc/Cargo.toml
```diff
@@ -8,6 +8,10 @@ homepage = "https://substrate.dev"
 repository = "https://github.com/paritytech/substrate/"
 description = "Substrate Client RPC"
 
+[features]
+rpc-disable-reserved-peers = []
```
jordy25519, line 13: can you add a comment here about the feature

jordy25519 (approved):
lgtm!

client/rpc/Cargo.toml
hoani: Done :)

mzxyz (approved):
Nice job, LGTM

PR #93, opened 2020-06-09
Add asset info to assets.

aliXsed:
* This info should be passed in when creating a new asset.
* Asset info should be updatable through an API.
* For the previously created assets, consider a default asset info.
* Check permissions before updating asset info.

### DIFF
--- a/frame/generic-asset/src/lib.rs
+++ b/frame/generic-asset/src/lib.rs
@@ -212,26 +212,23 @@ impl<T: Trait> Subtrait for T {
 /// Asset Metadata
 #[derive(Clone, PartialEq, Eq, Encode, Decode, Debug)]
 pub struct AssetInfo {
-    symbol: String,
-    decimal_places: u8,
+	symbol: String,
+	decimal_places: u8,
 }
 impl AssetInfo {
-    /// Create a new asset info by specifying its name/symbol and the number of decimal places
-    /// in the asset's balance. i.e. balance x 10 ^ -decimals will be the value for display
-    pub fn new(symbol: String, decimal_places: u8) -> Self {
-        Self {
-            symbol,
-            decimal_places,
-        }
-    }
+	/// Create a new asset info by specifying its name/symbol and the number of decimal places
+	/// in the asset's balance. i.e. balance x 10 ^ -decimals will be the value for display
+	pub fn new(symbol: String, decimal_places: u8) -> Self {
+		Self { symbol, decimal_places }
+	}
 }
 impl Default for AssetInfo {
-    fn default() -> Self {
-        Self {
-            symbol: String::default(),
-            decimal_places: 4,
-        }
-    }
+	fn default() -> Self {
+		Self {
+			symbol: String::default(),
+			decimal_places: 4,
+		}
+	}
 }
 
 /// Asset creation options.
@@ -476,11 +473,7 @@ decl_module! {
 		/// O(1) limited number of read and writes
 		/// Expected to not be called frequently
 		#[weight = SimpleDispatchInfo::FixedNormal(500_000)]
-		fn update_asset_info(
-			origin,
-			#[compact] asset_id: T::AssetId,
-			info: AssetInfo,
-		) -> DispatchResult {
+		fn update_asset_info(origin, #[compact] asset_id: T::AssetId, info: AssetInfo) -> DispatchResult {
 			let origin = ensure_signed(origin)?;
 
 			if !<TotalIssuance<T>>::contains_key(asset_id) {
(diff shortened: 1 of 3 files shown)


frame/generic-asset/src/lib.rs
```diff
@@ -209,6 +209,28 @@ impl<T: Trait> Subtrait for T {
 	type AssetId = T::AssetId;
 }
 
+/// Asset Metadata
+#[derive(Clone, PartialEq, Eq, Encode, Decode, Debug)]
+pub struct AssetInfo {
+    symbol: String,
+    decimal_points: u8,
```
jordy25519: `decimal_places` or `decimals` only 1 point :P

frame/generic-asset/src/lib.rs
```diff
@@ -526,6 +582,9 @@ decl_storage! {
 
 		/// The identity of the asset which is the one that is designated for paying the chain's transaction fee.
 		pub SpendingAssetId get(fn spending_asset_id) config(): T::AssetId;
+
+		/// The info for assets
+		pub AssetInfos: map hasher(twox_64_concat) T::AssetId => AssetInfo;
```
jordy25519: nit: maybe something like: `AssetProperties` / `AssetMeta` / `AssetData`

frame/generic-asset/src/lib.rs
```diff
@@ -209,6 +209,28 @@ impl<T: Trait> Subtrait for T {
...
 
+/// Asset Metadata
+#[derive(Clone, PartialEq, Eq, Encode, Decode, Debug)]
+pub struct AssetInfo {
+    symbol: String,
+    decimal_points: u8,
+}
+impl AssetInfo {
+	/// Create a new asset info by specifying its name/symbol and the designated number of
+	/// decimal points for the float representation of the balances
```
jordy25519: smth like: "the number of decimal places in the asset's balance. i.e. balance x 10 ^ -`decimals` will be the value for display."

frame/generic-asset/src/tests.rs
```diff
@@ -1196,3 +1196,95 @@ fn total_issuance_should_update_after_negative_imbalance_dropped() {
 			assert_eq!(GenericAsset::total_issuance(&asset_id), balance - 100);
 	});
 }
+
+#[test]
+fn asset_infos() {
+    ExtBuilder::default()
```
jordy25519: Please split the unit tests by public function atleast e.g. `create()` and `update_asset_info()` should be separate

KarishmaBothara (approved):

frame/generic-asset/src/lib.rs
```diff
@@ -948,13 +1011,25 @@ impl<T: Trait> Module<T> {
 		locks.retain(|l| l.id != id);
 		<Locks<T>>::insert(who, locks);
 	}
+
```
jordy25519: Doc comment / do we need this function if all assets are created with an `AssetInfo` 💭 (can always update storage once-off on CENNZnet for those that were created prior)

frame/generic-asset/src/lib.rs
jordy25519: also should be `config()`

frame/generic-asset/src/lib.rs
```diff
@@ -209,6 +209,28 @@ impl<T: Trait> Subtrait for T {
...
+	/// decimal points for the float representation of the balances
+	pub fn new(symbol: String, decimal_points: u8) -> Self {
+		Self { symbol, decimal_points, }
+	}
+}
+impl Default for AssetInfo {
+    fn default() -> Self {
+        Self {
+            symbol: String::default(),
+            decimal_points: 4,
```
jordy25519: maybe a runtime constant, since `4` is a CENNZnet decision

jordy25519 (commented):
lgtm overall!
some small naming nits and changes around configurability of the module

frame/generic-asset/src/lib.rs
aliXsed: Yeah :)

frame/generic-asset/src/lib.rs
alexsednz: I suggest we keep it. This gives the permission only to the owner of the asset and relieves us from adding temporary code or unnecessarily use Sudo. I will add doc comment.

frame/generic-asset/src/lib.rs
aliXsed: Anyway, here we need to have a default value in Plug, independent of the cennznet. As you said we have had the hardcoded assumption of 4, it makes everything to behave as before unless there's a specific configuration.

frame/generic-asset/src/lib.rs
aliXsed: Removed fn asset_info()

jordy25519 (approved):
lgtm!

PR #98, opened 2020-06-15
Added functions that set reserved_nodes

roy-powerfinance:
Added new function that sets reserved nodes in Peerset
Added a test

### DIFF
--- a/client/network/src/service.rs
+++ b/client/network/src/service.rs
@@ -678,7 +678,7 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkService<B, H> {
 		self.num_connected.load(Ordering::Relaxed)
 	}
 
-	pub fn set_reserved_nodes(&self, reserved_nodes:HashSet<PeerId> ) {
+	pub fn set_reserved_nodes(&self, reserved_nodes: HashSet<PeerId>) {
 		self.peerset.set_reserved_nodes(reserved_nodes);
 	}
 }
--- a/client/network/src/service.rs
+++ b/client/network/src/service.rs
@@ -677,6 +677,10 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkService<B, H> {
 	pub fn num_connected(&self) -> usize {
 		self.num_connected.load(Ordering::Relaxed)
 	}
+
+	pub fn set_reserved_nodes(&self, reserved_nodes:HashSet<PeerId> ) {
+		self.peerset.set_reserved_nodes(reserved_nodes);
+	}
 }
 
 impl<B: BlockT + 'static, H: ExHashT> sp_consensus::SyncOracle
--- a/client/peerset/src/lib.rs
+++ b/client/peerset/src/lib.rs
@@ -121,8 +121,8 @@ impl PeersetHandle {
 		let _ = self.tx.unbounded_send(Action::RemoveFromPriorityGroup(group_id, peer_id));
 	}
 
-	/// Remove a peer from a priority group.
-	pub fn set_reserved_nodes(&self, reserved_nodes:HashSet<PeerId>) {
+	/// Sets the "reserved" priority group
+	pub fn set_reserved_nodes(&self, reserved_nodes: HashSet<PeerId>) {
 		let _ = self.tx.unbounded_send(Action::SetReservedNodes(reserved_nodes));
 	}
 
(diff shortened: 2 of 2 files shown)


client/peerset/src/lib.rs
```diff
@@ -228,6 +235,19 @@ impl Peerset {
 		(peerset, handle)
 	}
 
+	pub fn set_reserved_nodes(&mut self, reserved_nodes : HashSet<PeerId>){
```
hoani: Should be following the `on_set_reserved_nodes` pattern since this is a dispatched call

client/network/src/service.rs
```diff
@@ -677,6 +677,10 @@ impl<B: BlockT + 'static, H: ExHashT> NetworkService<B, H> {
 	pub fn num_connected(&self) -> usize {
 		self.num_connected.load(Ordering::Relaxed)
 	}
+
+	pub fn set_reserved_nodes(&self, reserved_nodes:HashSet<PeerId> ) {
```
jordy25519: formatting: `reserved_nodes:HashSet<PeerId> )` -> `reserved_nodes: HashSet<PeerId>)`

client/peerset/src/lib.rs
```diff
@@ -119,6 +120,12 @@ impl PeersetHandle {
 	pub fn remove_from_priority_group(&self, group_id: String, peer_id: PeerId) {
 		let _ = self.tx.unbounded_send(Action::RemoveFromPriorityGroup(group_id, peer_id));
 	}
+
+	/// Remove a peer from a priority group.
+	pub fn set_reserved_nodes(&self, reserved_nodes:HashSet<PeerId>) {
```
jordy25519: `reserved_nodes: HashSet<PeerId>`, doc comment looks out of date

client/peerset/src/lib.rs
jordy25519: leading space: `reserved_nodes: HashSet<PeerId>`

client/peerset/src/lib.rs
```diff
@@ -570,6 +592,8 @@ mod tests {
 	use futures::prelude::*;
 	use super::{PeersetConfig, Peerset, Message, IncomingIndex, ReputationChange, BANNED_THRESHOLD};
 	use std::{pin::Pin, task::Poll, thread, time::Duration};
```
jordy25519: group imports: `use std::{collections::HashSet, pin::Pin, task::Poll, thread, time::Duration };`

client/peerset/src/lib.rs
```diff
@@ -610,6 +634,53 @@ mod tests {
 		]);
 	}
 
+
+	#[test]
+	fn test_peerset_set_reserved_peer() {
```
jordy25519: `test_peerset_set_reserved_nodes`

client/peerset/src/lib.rs
```diff
@@ -228,6 +235,19 @@ impl Peerset {
 		(peerset, handle)
 	}
 
+	pub fn set_reserved_nodes(&mut self, reserved_nodes : HashSet<PeerId>){
+		self.data.set_priority_group(RESERVED_NODES, reserved_nodes);
+
+		// If network is private, kick un-wanted connection off the network
+		if self.reserved_only {
+			self.on_set_reserved_only(true);
+		}
```
jordy25519, line 244: does it help to log something here if `self.reserved_only` is false?

jordy25519 (commented):
lgtm. few formatting nits to be addressed

client/peerset/src/lib.rs
roy-powerfinance: I think we expect this to be called periodically, so we may not want to spam the log

hoani: I don't know if this belongs here - it's not clear that there isn't a use-case where someone might want to explicitly set reserved nodes with `reserved_only = false`

client/peerset/src/lib.rs
```diff
@@ -610,6 +633,77 @@ mod tests {
 		]);
 	}
 
+
```
hoani, line 636: quite keen to see a case where we show that setting a reserved_peer list that already exists doesn't case a duplication in adds - is this possible?

client/peerset/src/lib.rs
roy-powerfinance: yeah it is true at the moment - My thought was that this new function simply provides the toolset to modify peerset, and it is up to the business logic layer to control when this function is used

client/peerset/src/lib.rs
roy-powerfinance: I found that hashset by definition does not allow duplication.
This applies to objects as well as primitives, I made a test project that verifies this as well

hoani: Cool, happy with that - the other tests kind of show it as well.

PR #99, opened 2020-06-16
Fix warnings

### LINKED ISSUE #97: Fix build warnings
Preparational step for releasing CENNZnet.

aliXsed:
Closes #97

### DIFF
--- a/frame/balances/src/mock.rs
+++ b/frame/balances/src/mock.rs
@@ -16,6 +16,9 @@
 
 //! Test utilities
 
+// TODO remove the following "allow" when merging in the latest version from upstream
+#![allow(dead_code)]
+
 use sp_runtime::{Perbill, traits::{ConvertInto, IdentityLookup}, testing::Header};
 use sp_core::H256;
 use sp_io;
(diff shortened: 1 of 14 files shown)


frame/balances/src/mock.rs
```diff
@@ -16,6 +16,9 @@
 
 //! Test utilities
 
+// TODO remove the following "allow" when merging in the latest version from upstream
+#![allow(dead_code)]
```
jordy25519, line 20: Can't say I'm a big fan of leaving TODOs like this.
What is "latest version" everytime or next time only?

jordy25519 (commented):
Not sure `#![allow(dead_code)]` is the right thing. Would prefer add a comment why some stuff is dead code or remove the dead code instead

aliXsed:
> Not sure `#![allow(dead_code)]` is the right thing. Would prefer add a comment why some stuff is dead code or remove the dead code instead

I have done that in places where there were several dead functions in the file with no signs of the code being removed in the upstream. So in those cases, I preferred to leave a very minimal footprint in the file to keep future upstream merges easy.

frame/balances/src/mock.rs
aliXsed: I have assumed the upstream warnings to be transitional. Anyone doing the next upstream merge would definitely come across this line and should make a decision about it. Ideally, with the next merge, we could remove the "allow" and that would be the preferred choice. Perhaps a better comment is needed here. One good practice with TODO comments that I know is there should be an issue created in the backlog and a reference to the issue number in the TODO comment, what's your opinion about this?

jordy25519 (approved):

PR #100, opened 2020-06-16
Fixed 2 broken tests

roy-powerfinance:
Fixed 2 failed tests:
* divide by zero
* host_function_not_found

### DIFF
--- a/primitives/arithmetic/src/fixed64.rs
+++ b/primitives/arithmetic/src/fixed64.rs
@@ -145,7 +145,7 @@ impl ops::Div for Fixed64 {
 
 	fn div(self, rhs: Self) -> Self::Output {
 		if rhs.0 == 0 {
-			panic!("divide by zero");
+			panic!("attempt to divide by zero");
 		}
 		let (n, d) = if rhs.0 < 0 {
 			(-self.0, rhs.0.abs() as u64)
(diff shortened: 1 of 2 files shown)


primitives/arithmetic/src/fixed64.rs
```diff
@@ -145,7 +145,7 @@ impl ops::Div for Fixed64 {
 
 	fn div(self, rhs: Self) -> Self::Output {
 		if rhs.0 == 0 {
-			panic!("divide by zero");
+			panic!("attempt to divide by zero");
```
aliXsed: If the failure is because of expecting a particular message in a test, you'd rather change the expectation, not the code as nothing is wrong with the code.

roy-powerfinance: yeah good point alex

aliXsed (approved):

primitives/arithmetic/src/fixed64.rs
hoani: Is this change needed to make the tests pass again? If not, let's just make the test pass

hoani (approved):
lgtm, thanks for taking care of this @royyang-centrality !

aliXsed:
Interestingly for me with the following toolchain which is the same as this project's CircleCi config
`nightly-2020-04-17-x86_64-apple-darwin
1.42.0-x86_64-apple-darwin`
there is no failures for these tests.
@royyang-centrality what is your toolchain setup?

roy-powerfinance:
@alexsednz I think i'm using a later version of the compiler. 
I find it strange as well

PR #101, opened 2020-06-16
Fixed some problem where test won't compile or fail

roy-powerfinance:
Fixed some test compile problem and failed tests

### DIFF
--- a/frame/support/src/dispatch.rs
+++ b/frame/support/src/dispatch.rs
@@ -179,7 +179,7 @@ impl<T> Parameter for T where T: Codec + EncodeLike + Clone + Eq + Any + fmt::De
 /// pub trait Trait: system::Trait where Self::AccountId: From<u32> {}
 ///
 /// decl_module! {
-/// 	pub struct Module<T: Trait> for enum Call where origin: T::Origin, <T as frame_system::Trait>::AccountId: From<u32> {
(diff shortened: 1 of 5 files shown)


frame/support/src/dispatch.rs
```diff
@@ -2075,7 +2075,7 @@ mod tests {
 		use super::*;
 
 		pub trait Trait {
-			type AccountId;
+			type AccountId: From<u32>;
```
aliXsed, line 2078: With this change, you'd better remove the "where" constraint from:
`pub trait Trait: system::Trait + Sized where Self::AccountId: From<u32>`

frame/support/src/weights.rs
```diff
@@ -335,7 +335,7 @@ mod tests {
 	mod system {
 		use super::*;
 		pub trait Trait {
-			type AccountId;
+			type AccountId: From<u32>;
```
aliXsed, line 338: Same here.

aliXsed:
@royyang-centrality regarding the update of dependencies in Cargo.lock, if you haven't intended them, I suggest reverting. I'm thinking perhaps we'd better update dependencies right after a release rather than close to a release.

frame/support/src/dispatch.rs
roy-powerfinance: Yep will fix

aliXsed (approved):

PR #109, opened 2020-07-06
Add has_keys to session module

hoani:
We want to be able to check with the session module whether a validator actually has its session keys configured before updating it with a new set of validators.

## Changes

* Add function `has_keys` to check if an account has associated keys added to the `session` module
* Some tests for `has_keys`

## Concerns

* None

### DIFF
--- a/frame/session/src/lib.rs
+++ b/frame/session/src/lib.rs
@@ -697,7 +697,7 @@ impl<T: Trait> Module<T> {
 	}
 
 	pub fn has_keys(v: &T::ValidatorId) -> bool {
-		<NextKeys<T>>::get(v).is_some()
+		<NextKeys<T>>::contains_key(v)
 	}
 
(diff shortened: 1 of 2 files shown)


frame/session/src/lib.rs
```diff
@@ -696,6 +696,10 @@ impl<T: Trait> Module<T> {
 		<NextKeys<T>>::take(v)
 	}
 
+	pub fn has_keys(v: &T::ValidatorId) -> bool {
+		<NextKeys<T>>::get(v).is_some()
```
jordy25519: nit: I think `::contains_key(v)` is more efficient for a check as it doesn't deserialize the value

jordy25519 (approved):
lgtm

cowboy-bebug (approved):
LGTM!

PR #123, opened 2020-08-18
Migrated Consortium permission module into Plug

roy-powerfinance:
Add prml consortium permission module into Plug


cowboy-bebug (approved):
lgtm!

PR #126, opened 2020-08-25
Tidy up docker

cowboy-bebug:
Here are some tidy ups around docker to help streamline future testing.

Changes:
- Create and push with `latest` tag when `publish-docker` job is triggered
- Copy `.maintain/Dockerfile` into `.plug/Dockerfile` to maintain plug-specific Dockerfile
- Change `CMD` -> `ENTRYPOINT` in the Dockerfile so that parameters can be passed
- Export `TERM=xterm` for the case when `$TERM` isn't set (such as in docker)
- Add prometheus and grafana services (dashboards not yet created)

### DIFF
--- a/.plug/logger.sh
+++ b/.plug/logger.sh
@@ -4,6 +4,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+export TERM=xterm
+
 bold=$(tput bold)
 red=$(tput setaf 1)
 green=$(tput setaf 2)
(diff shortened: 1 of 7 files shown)


.plug/logger.sh
```diff
@@ -4,6 +4,8 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
+export TERM=xterm
```
jordy25519:
```suggestion
export TERM=${TERM:-xterm}
```

jordy25519 (approved):
Thanks for this, it lgtm!

PR #127, opened 2020-08-26
Added the IssuerPermissions back to solve an bug

roy-powerfinance:
Resolves the problem where issuers don't have access to send extrinsics to give self the permission to send extrinsics.
Added some unit tests back

### DIFF
--- a/prml/consortium-permission/src/lib.rs
+++ b/prml/consortium-permission/src/lib.rs
@@ -296,7 +296,9 @@ impl<T: Trait> Module<T> {
     fn initialise_issuers(issuers: &Vec<(T::AccountId, Vec<Topic>)>) {
         for (issuer, topics) in issuers {
             Issuers::<T>::insert(issuer, topics);
-            T::IssuerPermissions::grant_issuer_permissions(&issuer, &topic);
+            for topic in topics {
+                T::IssuerPermissions::grant_issuer_permissions(&issuer, &topic);
+            }
         }
     }
 
--- a/prml/consortium-permission/src/lib.rs
+++ b/prml/consortium-permission/src/lib.rs
@@ -72,6 +72,18 @@ pub type Topic = Vec<u8>;
 /// Type used for values of corresponding topics.
 pub type Value = Vec<u8>;
 
+/// Allows runtime implmentation of issuer configuration.
+pub trait IssuerPermissions {
+    type AccountId;
+    type Topic;
+
+    /// Grants the permission to issue claims for a new topic.
+    fn grant_issuer_permissions(issuer: &Self::AccountId, topic: &Topic);
+
+    /// Revokes the permission to issue claims for a topic.
+    fn revoke_issuer_permissions(issuer: &Self::AccountId, topic: &Topic);
+}
+
 /// The module's config trait.
 pub trait Trait: frame_system::Trait {
     /// The overarching event type.
@@ -80,6 +92,8 @@ pub trait Trait: frame_system::Trait {
     type MaximumTopicSize: Get<usize>;
     /// The maximum number of bytes allowed for a value.
     type MaximumValueSize: Get<usize>;
+    /// Provides an interface for setting issuer permissions
+    type IssuerPermissions: IssuerPermissions<AccountId = <Self as frame_system::Trait>::AccountId, Topic = Topic>;
 }
 
 decl_storage! {
@@ -171,6 +185,8 @@ decl_module! {
             current_topics.push(topic.clone());
             Issuers::<T>::insert(who.clone(), current_topics);
 
+            T::IssuerPermissions::grant_issuer_permissions(&who, &topic);
+
             Self::deposit_event(RawEvent::IssuerWithTopicAdded(who, topic));
         }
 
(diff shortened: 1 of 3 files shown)


cowboy-bebug (approved):
LGTM

prml/consortium-permission/src/mock.rs
```diff
@@ -36,6 +36,33 @@ pub const ACCESS_TOPIC: &[u8; 6] = b"access";
...
+        if *topic == ACCESS_TOPIC {
+            ConsortiumPermission::do_make_claim(
+                issuer,
+                issuer,
+                ACCESS_TOPIC.to_vec().as_ref(),
+                vec![ACCESS_VALUE].as_ref(),
+            );
+        }
+    }
+    /// Remove all self-claimed permissions from an issuer.
```
jordy25519: Not sure if comment wrong or implementation is wrong (maybe both)

/// Remove all self-claimed permissions from an issuer.
or
/// Remove self-claimed access permission for`issuer` (it does this afaict)

does it need the if block, since `ConsortiumPermission::claim((issuer, ACCESS_TOPIC.to_vec()));` would return `ACCESS_VALUE` false / unset if it doesn't exist

jordy25519 (changes requested):
revoking part needs some reworking

prml/consortium-permission/src/mock.rs
roy-powerfinance: Roger. I'll update the comments and double check this

prml/consortium-permission/src/lib.rs
```diff
@@ -203,6 +221,14 @@ decl_module! {
         /// Requires Root.
         pub fn force_remove_issuer(origin, who: T::AccountId) {
             ensure_root(origin)?;
+
+            // Notify the revocation of all current permissions.
```
jordy25519, line 225: this comment doesn't line up with the implementation of `revoke_issuer_permissions`

roy-powerfinance: revoke_issuer_permissions can behave differently depending on its implementation. 
the module will simply notify the revocation by calling the revoke_issuer_permissions on all the topics it is revoking

jordy25519: makes sense!

prml/consortium-permission/src/mock.rs
roy-powerfinance: I think the if block is needed. Because the "access" can be granted by another person. We should only revoke "access" if it is granted by the "issuer" themselves when "access" is revoked

roy-powerfinance: * Updated the comments here.

prml/consortium-permission/src/mock.rs
```diff
@@ -36,6 +36,34 @@ pub const ACCESS_TOPIC: &[u8; 6] = b"access";
 /// Reserved value for access to submit an extrinsic.
 pub const ACCESS_VALUE: u8 = 1;
 
+pub struct IssuerPermissionsMock;
+
+impl IssuerPermissions for IssuerPermissionsMock {
+    type AccountId = AccountId;
+    type Topic = Topic;
+    /// When an issuer is authorized to make claims on the "access" topic, also grant them the
+    /// "access" permission.
```
jordy25519:
```suggestion
    /// When an issuer is authorized to make claims on the "access" topic, they also grant themselves
    /// "access" permission.
```

prml/consortium-permission/src/mock.rs
```diff
@@ -36,6 +36,34 @@ pub const ACCESS_TOPIC: &[u8; 6] = b"access";
...
+            ConsortiumPermission::do_make_claim(
+                issuer,
+                issuer,
+                ACCESS_TOPIC.to_vec().as_ref(),
+                vec![ACCESS_VALUE].as_ref(),
+            );
+        }
+    }
+    /// When an issuer's authority on the "access" topic is revoked, also revoke their "access" permission.
+    fn revoke_issuer_permissions(issuer: &Self::AccountId, topic: &Topic) {
```
jordy25519:
```suggestion
    /// When an issuer's authority on the "access" topic is revoked, also revoke their self-claimed "access" permission.
    fn revoke_issuer_permissions(issuer: &Self::AccountId, topic: &Topic) {
```

jordy25519 (commented):
looking good now
- mentioning self-claim is an important quirk imo
- missing one test for when `claim_issuer != issuer` (i.e. hit that if block)

prml/consortium-permission/src/mock.rs
jordy25519: gotcha. I meant the outer if, as this function seemed made to remove self access claim but now I see it _could_ be called with any topic

prml/consortium-permission/src/mock.rs
jordy25519: don't commit this not sure if whitespace is correct 😅

prml/consortium-permission/src/mock.rs
roy-powerfinance: I want to make it so that we only revoke the "access" permission if when we revoke their authority on the "access" topic.
Because we could be revoking their rights on other topics, such as mint_burn, and others, but still want them to keep their "access" right. Otherwise revoking any authority will revoke their "access" right, which seems very harsh
does that make sense?

roy-powerfinance:
> looking good now
> 
> * mentioning self-claim is an important quirk imo
> * missing one test for when `claim_issuer != issuer` (i.e. hit that if block)

Yeah good point - I've fixed up the comments and added some more unit tests

jordy25519 (approved):
nice one

PR #155, opened 2020-10-27
Move common traits into prml-common

jordy25519:


prml/common/Cargo.toml
```diff
@@ -0,0 +1,27 @@
+[package]
+name = "prml-common"
```
aliXsed: Can we name it prml-multi-currency?

aliXsed (approved):

PR #162, opened 2020-11-18
Fix compilation with latest nightly

jordy25519:
Cherry-picks: make Substrate compile with latest nightly (#7381)


jordy25519:
decided it would be simpler to get the build working on latest nightly by itself :D
cc #161

PR #163, opened 2020-12-09
Add custom decode to support RuntimeDispatchInfo when Weight=u32

jordy25519:
As it says on the tin ☝️
Not as extensible as something deeper but enough to support the upgrade process from pre-2.0 to post-2.0

-- 

Will test in CENNZnet cc https://github.com/cennznet/cennznet/issues/341

### DIFF
--- a/frame/transaction-payment/rpc/runtime-api/src/lib.rs
+++ b/frame/transaction-payment/rpc/runtime-api/src/lib.rs
@@ -27,7 +27,7 @@ use serde::{Serialize, Deserialize, Serializer, Deserializer};
 use sp_runtime::traits::{MaybeDisplay, MaybeFromStr};
 
 /// Information related to a dispatchable's class, weight, and fee that can be queried from the runtime.
-#[derive(Eq, PartialEq, Encode, Decode, Default)]
+#[derive(Eq, PartialEq, Encode, Default)]
 #[cfg_attr(feature = "std", derive(Debug, Serialize, Deserialize))]
 #[cfg_attr(feature = "std", serde(rename_all = "camelCase"))]
 pub struct RuntimeDispatchInfo<Balance> {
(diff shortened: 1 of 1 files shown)


frame/transaction-payment/rpc/runtime-api/src/lib.rs
```diff
@@ -44,6 +44,38 @@ pub struct RuntimeDispatchInfo<Balance> {
...
+const LEGACY_RUNTIME_DISPATCH_INFO_BYTE_LENGTH: usize = 21;
+
+impl<Balance: Decode> Decode for RuntimeDispatchInfo<Balance> {
+	// Custom decode implementation to handle the differences between the `RuntimeDispatchInfo` type
+	// between client version vs. runtime version
+	// Concretely, `Weight` type changed from `u32` in some legacy runtimes to now `u64`
+	fn decode<I: codec::Input>(value: &mut I) -> Result<Self, codec::Error> {
+		// Check `value` len to see whether we should decode legacy or new Weight type
+		let input_len = value.remaining_len()?.ok_or("empty buffer while decoding")?;
+		let weight: Weight = if input_len == LEGACY_RUNTIME_DISPATCH_INFO_BYTE_LENGTH {
```
aliXsed, line 62: Perhaps it could be safer if we use std::mem::size_of for determining the length of the native RuntimeDispatchInfo and if it was accidentally the same as LEGACY_RUNTIME_DISPATCH_INFO_BYTE_LENGTH, we still stick to normal decoding instead of the legacy one.

jordy25519: Interesting, I tried it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=4f86b71ba1ed84495f6879fe99742a22 I don't think the memory layout of the struct is the same as the scale encoded byte length

jordy25519:
something wrong with CI build but tests passing 🤷‍♀️

frame/transaction-payment/rpc/runtime-api/src/lib.rs
aliXsed: I checked using #[repr(C)] in your example but it also returned the same result as yours.

PR #168, opened 2021-01-20
RPC Restart Grandpa Voter

jordy25519:
Adds an RPC to trigger a "restart" of the grandpa voter
This allows the node to use new grandpa session/authority keys without a full node restart or waiting for authority set change event.

Steps to reproduce:
1) Start a local node `--chain=dev --validator --rpc-method=Unsafe`
2) Insert session keys for babe and gran
3) trigger the worker restart: `curl -H 'Content-Type: application/json' -d '{ "jsonrpc": "2.0", "method":"grandpa_restartVoter", "params":[], "id": 1 }' http://localhost:9933`

Screen Shot 2021-01-21 at 11 53 12

cc: https://github.com/paritytech/substrate/issues/7268

### DIFF
--- a/client/finality-grandpa/src/lib.rs
+++ b/client/finality-grandpa/src/lib.rs
@@ -375,20 +375,22 @@ impl<Block, Network> BlockSyncRequester<Block> for NetworkBridge<Block, Network>
 
 /// A new authority set along with the canonical block it changed at.
 #[derive(Debug)]
-pub(crate) struct NewAuthoritySet<H, N> {
-	pub(crate) canon_number: N,
-	pub(crate) canon_hash: H,
-	pub(crate) set_id: SetId,
-	pub(crate) authorities: AuthorityList,
+pub struct NewAuthoritySet<H, N> {
+	pub canon_number: N,
+	pub canon_hash: H,
+	pub set_id: SetId,
+	pub authorities: AuthorityList,
 }
 
 /// Commands issued to the voter.
 #[derive(Debug)]
-pub(crate) enum VoterCommand<H, N> {
+pub enum VoterCommand<H, N> {
 	/// Pause the voter for given reason.
 	Pause(String),
 	/// New authorities.
-	ChangeAuthorities(NewAuthoritySet<H, N>)
+	ChangeAuthorities(NewAuthoritySet<H, N>),
+	/// Restart the local grandpa worker
+	Restart,
 }
 
 impl<H, N> fmt::Display for VoterCommand<H, N> {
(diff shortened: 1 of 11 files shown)


client/finality-grandpa/src/lib.rs
```diff
@@ -375,27 +375,34 @@ impl<Block, Network> BlockSyncRequester<Block> for NetworkBridge<Block, Network>
 
 /// A new authority set along with the canonical block it changed at.
 #[derive(Debug)]
-pub(crate) struct NewAuthoritySet<H, N> {
-	pub(crate) canon_number: N,
-	pub(crate) canon_hash: H,
-	pub(crate) set_id: SetId,
-	pub(crate) authorities: AuthorityList,
+pub struct NewAuthoritySet<H, N> {
```
aliXsed, line 378: Is changing pub(crate) to pub necessary?

client/finality-grandpa/src/lib.rs
```diff
@@ -375,27 +375,34 @@ impl<Block, Network> BlockSyncRequester<Block> for NetworkBridge<Block, Network>
...
+	/// Authority set id
+	pub set_id: SetId,
+	/// The authority IDs
+	pub authorities: AuthorityList,
 }
 
 /// Commands issued to the voter.
 #[derive(Debug)]
-pub(crate) enum VoterCommand<H, N> {
+pub enum VoterCommand<H, N> {
```
aliXsed, line 391: Same here

jordy25519: Yeah I wanted to issue a `VoterCommand` to the grandpa voter from the rpc code so this type became used outside the sc-finality-grandpa crate

client/finality-grandpa/src/lib.rs
jordy25519: this one was a necessary flow on from making `VoterCommand` pub

aliXsed:
Do we know what is the value of `voter_set_state` when a grandpa authority change (the condition where we need to send the restart command) has happened?

jordy25519:
> Do we know what is the value of `voter_set_state` when a grandpa authority change (the condition where we need to send the restart command) has happened?

My understanding is, every node always independently tracks the grandpa round state / votes etc. so we don't need any special handling.
The problem this PR fixes is that the grandpa voter worker is not aware it should be voting when it is newly given a session key that should be voting.

I think this is due to the value of the node's session key being cached in-memory somehow, the value is not properly refreshed until `rebuild_voter` happens which only happens on AuthoritySet change event or first start.
This PR allows the the refresh to happen on demand, the grandpa voter state is not altered in anyway by this PR, simply rebuilt with current available state.

aliXsed:
LGTM! though it's hard to track a potential side effect of rebuilding the grandpa state. But given that you have tested it, it should be safe. Also we know that this rpc is an unsafe one and wouldn't be exposed publicly.

aliXsed (approved):

PR #179, opened 2021-03-03
latest upstream merge

aliXsed:
Brings the latest upstream changes and fix our prml modules "generic asset" and "attestation" to compile and test correctly.


jordy25519 (commented):
is it necessary to remove the codec trait bound? I guess it will cause problems for runtime modules which use this trait e.g. cennzx

https://github.com/plugblockchain/plug-blockchain/pull/179/files#diff-4006b49434d54edb0558ad94b80bd27bdb6a7d738e3c8026be60e48013690cf8L25-L50

apart from that looks good

aliXsed:
> is it necessary to remove the codec trait bound? I guess it will cause problems for runtime modules which use this trait e.g. cennzx
> 
> https://github.com/plugblockchain/plug-blockchain/pull/179/files#diff-4006b49434d54edb0558ad94b80bd27bdb6a7d738e3c8026be60e48013690cf8L25-L50
> 
> apart from that looks good

Yes it was necessary. The way it was before, MultiCurrencyAccounting had types bound to codec. Then in the generic implementation of it for the generic-asset module, the config's associated types would be assigned to those of MultiCurrencyAccounting. It was while the Config's types had less bounds and weren't bound to codec. That was the reason the latest stable Rust was unhappy when compiling it.  I tried to restrict the implementation by introducing a where clause to solve it. However, I couldn't get it working perhaps due to the fact in our case there are several layers of traits in the hierarchy. For example, the type AccountId coming down from the frame_system::Config to primal_generic_asset::Config and then would be referenced in the implementation of MultiCurrencyAccounting for primal_generic_asset::Module<T>. Setting a where clause to restrict AccountId there wasn't an easy job for the compiler.

jordy25519:
Ok I recommend trying to build cennzx with this branch to see if it causes any issues, otherwise we're good

aliXsed:
> Ok I recommend trying to build cennzx with this branch to see if it causes any issues, otherwise we're good

@holygits I could successfully and easily build cennzx by using this branch. The only change that I had to make was:
Screen Shot 2021-03-04 at 6 28 26 PM
Then replacing BalanceOf<T> with T::Balance everywhere in that code. IMO the resulting change makes the code read better by avoiding BalanceOf.

jordy25519 (approved):

Issue #181, opened 2021-03-11
Leave account storage handling in Generic Asset to an account store


aliXsed:
The study of recent upstream changes suggests to me that our generic asset module which is replacing the balances pallet should use an account store when dealing with creating or removing accounts. An account store will be bound to `StoredMap` and its functionalities is already implemented by `frame_system::Module`. The generic asset should only call functions like `insert` or `try_mutate_exists` when appropriate.

jordy25519:
an update on this, the account store doesn't need to track balances at this stage but we have the API to support account existence checks in wider system/frame context

PR #182, opened 2021-03-12
Use a StoredMap to indicate the existence of generic asset accounts 

aliXsed:
This minimally addresses #181. 
The reason for removing generic asset from Plug's runtime in this PR is that pallet_balances there overlaps with pallet_generic_asset and causes a conflict in this functionality. However pallet_generic_asset in its current implmentation cannot fully replace pallet_balances. As we had added the generic asset to Plug's runtime only for the sake of benchmarking, we can let cennznet do the benchmarking as in CENNZnet generic_asset works okay with all the other modules there.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4146,7 +4146,6 @@ dependencies = [
  "pallet-vesting",
  "parity-scale-codec",
  "prml-attestation",
- "prml-generic-asset",
  "serde",
  "sp-api",
  "sp-authority-discovery",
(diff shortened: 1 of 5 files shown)


jordy25519 (commented):
Operations like transfer can make accounts exist too
We'd need to consider clean up mechanisms to make accounts 'not exist' also aka existential deposit rules. problem is you might need to check the free balance of all existing assets to know if the account "exists" or not.
this needs more design work

aliXsed:
> Operations like transfer can make accounts exist too
> We'd need to consider clean up mechanisms to make accounts 'not exist' also aka existential deposit rules. problem is you might need to check the free balance of all existing assets to know if the account "exists" or not.
> this needs more design work

Generic Asset is not prodving the feature for removing accounts when their balance goes below the existential deposit. This PR is based on what we have there already. I can work on that in an individual PR right after this one. It keeps the change as small and coherent as possible which we have always longed for.

jordy25519 (approved):

KarishmaBothara (approved):

aliXsed:
As discussed, the solution here and the interaction with the frame_system will become complete after some other upcoming PRs that we are going to have for #183

Issue #183, opened 2021-03-16
Purge generic asset accounts with balances less than existential deposit 


aliXsed:
Asset info for each asset should feature the existential deposit for that asset. Generic Asset, post any operations that would potentially leave dust asset for an account should update the account data in the account store and purge the insignificant asset if the policy of that operation allows it. The account would be removed from the account store as well when there is no significant assets left. 
A migration logic would be needed so the current storage of the accounts is searched for dust balances.
Endowed accounts defined in the genesis will be exempt from the dust balance check. Or in other worlds they should persist anyway.

PR #184, opened 2021-03-16
Add asset info extra field for existential_deposit

aliXsed:
First step for #183, before we can trace accounts with dust values.

### DIFF
--- a/prml/generic-asset/rpc/runtime-api/src/lib.rs
+++ b/prml/generic-asset/rpc/runtime-api/src/lib.rs
@@ -30,9 +30,9 @@ use sp_std::vec::Vec;
 
 sp_api::decl_runtime_apis! {
 	/// The API to query asset meta information.
-	pub trait AssetMetaApi<AssetId: Decode + Encode>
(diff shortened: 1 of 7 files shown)


jordy25519 (commented):
PR looks good to me! I think the existential deposit should never be allowed as `0`, maybe it can be some function of decimal places by default or atleast `1`

aliXsed:
> PR looks good to me! I think the existential deposit should never be allowed as `0`, maybe it can be some function of decimal places by default or atleast `1`

If I force a non-zero setting, I'm afraid it will become very opinionated getting in the way of some possible future use cases of assets for which the creator believes 0 is okay and the accounts should exist all the time. What do you think? @holygits

jordy25519:
I think ED is required as a sensible network policy not a per asset policy. We could make the module configurable as such but probably just over complicates this right now, we should build for the uses cases we have right now not the ones we might have

jordy25519 (approved):
lgtm!

KarishmaBothara (approved):

Issue #185, opened 2021-03-18
Place the locks for a specifed asset


aliXsed:
Currently when a lock is placed for an account, it will be used against all the assets (different asset ids) of that account which is wrong. The locks should be keyed by the asset ids and checked against the same asset id for which they are placed. 
We may then need to implement lockable currency not for the whole generic asset module, but for a currency that is maintained by generic asset. Currencies like spending asset or staking asset and so on.

PR #186, opened 2021-03-22
Implement existential deposit for generic asset

### LINKED ISSUE #183: Purge generic asset accounts with balances less than existential deposit

aliXsed:
Closes #183. Keeps trace of significant assets for account ids and claim dust balances if the policy says so. Also add a migration logic for updating the newly added account store from the current storage and purging dust balances on runtime upgrade.

### DIFF
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -815,6 +815,7 @@ impl<T: Config> Module<T> {
 		AssetMeta::<T>::iter().collect()
 	}
 
+	/// Return true if the specified asset of `who` is considered dust (insignificant).
 	fn is_dust(asset_id: T::AssetId, who: &T::AccountId) -> bool {
 		let existential_deposit = AssetMeta::<T>::get(asset_id).existential_deposit();
 		// If for an asset, there is enough deposit above the defined existential deposit, it will not
@@ -825,6 +826,7 @@ impl<T: Config> Module<T> {
 			&& Self::locks(who).is_empty()
 	}
 
+	/// Try to mutate account of `who` in the account store based on `is_dust` for the specified asset.
 	fn try_mutate_account(asset_id: T::AssetId, who: &T::AccountId, is_dust: bool) -> Option<AccountData<T::AssetId>> {
 		T::AccountStore::try_mutate_exists(who, |maybe_account| {
 			match maybe_account {
@@ -863,7 +865,7 @@ impl<T: Config> Module<T> {
 	}
 
 	/// Mutate an account to some new value, or delete it entirely according to existence requirement
-	/// Return true if the account is mutated or false if it's deleted and thus got a zero balance.
+	/// Return true if the account is mutated or false if it's deleted.
 	fn try_mutate_account_with_dust(
 		asset_id: T::AssetId,
 		who: &T::AccountId,
@@ -892,6 +894,7 @@ impl<T: Config> Module<T> {
 		true
 	}
 
+	/// Set both `free_balance` and `reserved_balance` in one go meaning updating the account store once
 	fn set_balances(asset_id: T::AssetId, who: &T::AccountId, free_balance: T::Balance, reserved_balance: T::Balance) {
 		let _ = Self::try_mutate_account_with_dust(
 			asset_id,
@@ -904,8 +907,8 @@ impl<T: Config> Module<T> {
 		);
 	}
 
-	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance. It is expected that
-	/// the caller will do this.
+	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance unless the account is
+	/// purged in which case it would only create and drop the imbalances for the purged stage.
 	fn set_reserved_balance(asset_id: T::AssetId, who: &T::AccountId, balance: T::Balance) {
 		let _ = Self::try_mutate_account_with_dust(
 			asset_id,
@@ -917,8 +920,8 @@ impl<T: Config> Module<T> {
 		);
 	}
 
-	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance. It is expected that
-	/// the caller will do this.
+	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance. unless the account is
+	/// purged in which case it would only create and drop the imbalances for the purged stage.
 	fn set_free_balance(asset_id: T::AssetId, who: &T::AccountId, free_balance: T::Balance) {
 		let _ = Self::try_mutate_account_with_dust(
 			asset_id,
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -190,6 +190,7 @@ use weights::WeightInfo;
 pub struct AccountData<AssetId: Ord> {
 	/// Contains current existing asset ids for an account
 	existing_assets: BTreeSet<AssetId>,
+	exists: bool,
 }
 
 impl<AssetId: AtLeast32BitUnsigned + Copy> AccountData<AssetId> {
@@ -833,20 +834,13 @@ impl<T: Config> Module<T> {
 	/// Update the account of `who` in the account store based on the current asset status. Pass
 	/// true in `asset_is_dust`, if the asset for `who` is at the time of this call considered dust.
 	/// This will purge dust assets.
-	fn update_account_store(
-		asset_id: T::AssetId,
-		who: &T::AccountId,
-		asset_is_dust: bool,
-	) -> Option<AccountData<T::AssetId>> {
-		T::AccountStore::try_mutate_exists(who, |maybe_account| {
+	fn update_account_store(asset_id: T::AssetId, who: &T::AccountId, asset_is_dust: bool) {
+		let _ = T::AccountStore::try_mutate_exists(who, |maybe_account| {
 			match maybe_account {
 				Some(account) => {
 					if asset_is_dust {
 						account.existing_assets.remove(&asset_id);
 						Self::purge(asset_id, who);
-						if !account.should_exist() {
-							*maybe_account = None;
-						}
 					} else {
 						account.existing_assets.insert(asset_id);
 					}
@@ -855,13 +849,19 @@ impl<T: Config> Module<T> {
 					if !asset_is_dust {
 						let mut account: AccountData<T::AssetId> = Default::default();
 						account.existing_assets.insert(asset_id);
+						account.exists = true;
 						*maybe_account = Some(account);
 					}
 				}
 			};
 			<Result<_, &'static str>>::Ok(maybe_account.clone())
-		})
-		.unwrap_or_default()
+		});
+
+		if asset_is_dust && !T::AccountStore::get(who).should_exist() {
+			// This is just an attempt to remove the account.
+			// Account store might still keep it if there are consumers for this account.
+			let _ = T::AccountStore::remove(who);
+		}
 	}
 
 	/// Remove an asset for an account and pass a non-zero imbalance to dust imbalance handler.
@@ -883,7 +883,7 @@ impl<T: Config> Module<T> {
 		let is_dust = Self::is_dust(asset_id, who);
 
 		if (is_dust && req == ExistenceRequirement::AllowDeath) || (!is_dust && was_dust) {
-			let _ = Self::update_account_store(asset_id, who, is_dust);
+			Self::update_account_store(asset_id, who, is_dust);
 		}
 	}
 
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -220,7 +220,7 @@ pub trait Config: frame_system::Config {
 	/// The system event type
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
 
-	/// The means of storing the balances of an account.
+	/// The means of storing account data for account ids.
 	type AccountStore: StoredMap<Self::AccountId, AccountData<Self::AssetId>>;
 
 	/// The treasury account for clearing up dusts.
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -195,6 +195,7 @@ pub struct AccountData<AssetId: Ord> {
 
 impl<AssetId: AtLeast32BitUnsigned + Copy> AccountData<AssetId> {
 	/// Return true if the account should be kept in the account store.
+	#[allow(dead_code)]
 	fn should_exist(&self) -> bool {
 		!self.existing_assets.is_empty()
 	}
@@ -851,7 +852,7 @@ impl<T: Config> Module<T> {
 		// be considered a dust asset. Also any reservation or locks on the asset would mean the asset
 		// should be kept for the clearance of those operations and thus is not dust.
 		<FreeBalance<T>>::get(asset_id, who) < existential_deposit
-			&& <ReservedBalance<T>>::get(asset_id, who) == Zero::zero()
+			&& <ReservedBalance<T>>::get(asset_id, who).is_zero()
 			&& Self::locks(who).is_empty()
 	}
 
@@ -880,11 +881,12 @@ impl<T: Config> Module<T> {
 			<Result<_, &'static str>>::Ok(maybe_account.clone())
 		});
 
-		if asset_is_dust && !T::AccountStore::get(who).should_exist() {
-			// This is just an attempt to remove the account.
-			// Account store might still keep it if there are consumers for this account.
-			let _ = T::AccountStore::remove(who);
-		}
+		// TODO Enable the following logic after https://github.com/plugblockchain/plug-blockchain/issues/191
+		// if asset_is_dust && !T::AccountStore::get(who).should_exist() {
+		// 	// This is just an attempt to remove the account.
+		// 	// Account store might still keep it if there are consumers for this account.
+		// 	let _ = T::AccountStore::remove(who);
+		// }
 	}
 
 	/// Remove an asset for an account and pass a non-zero imbalance to dust imbalance handler.
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -914,8 +914,8 @@ impl<T: Config> Module<T> {
 		);
 	}
 
-	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance unless the account is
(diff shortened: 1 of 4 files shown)


prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +815,122 @@ impl<T: Config> Module<T> {
...
 
-	// PRIVATE MUTABLES
+	/// Return true if the specified asset of `who` is considered dust (insignificant).
+	fn is_dust(asset_id: T::AssetId, who: &T::AccountId) -> bool {
+		let existential_deposit = AssetMeta::<T>::get(asset_id).existential_deposit();
+		// If for an asset, there is enough deposit above the defined existential deposit, it will not
+		// be considered a dust asset. Also any reservation or locks on the asset would mean the asset
+		// should be kept for the clearance of those operations and thus is not dust.
+		<FreeBalance<T>>::get(asset_id, who) < existential_deposit
+			&& <ReservedBalance<T>>::get(asset_id, who) == Zero::zero()
```
jordy25519: can reserve balance be below ED and does that matter 💭

prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +815,122 @@ impl<T: Config> Module<T> {
...
+			&& Self::locks(who).is_empty()
+	}
 
-	/// NOTE: LOW-LEVEL: This will not attempt to maintain total issuance. It is expected that
-	/// the caller will do this.
+	/// Try to mutate account of `who` in the account store based on `is_dust` for the specified asset.
+	fn try_mutate_account(asset_id: T::AssetId, who: &T::AccountId, is_dust: bool) -> Option<AccountData<T::AssetId>> {
```
jordy25519: nit: it's not clear what `is_dust` means in this context
would also rename function to describe what it actually does e.g. `create_or_update_account_assets`

prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +815,122 @@ impl<T: Config> Module<T> {
...
-	// PRIVATE MUTABLES
+	/// Return true if the specified asset of `who` is considered dust (insignificant).
+	fn is_dust(asset_id: T::AssetId, who: &T::AccountId) -> bool {
+		let existential_deposit = AssetMeta::<T>::get(asset_id).existential_deposit();
+		// If for an asset, there is enough deposit above the defined existential deposit, it will not
+		// be considered a dust asset. Also any reservation or locks on the asset would mean the asset
+		// should be kept for the clearance of those operations and thus is not dust.
+		<FreeBalance<T>>::get(asset_id, who) < existential_deposit
+			&& <ReservedBalance<T>>::get(asset_id, who) == Zero::zero()
+			&& Self::locks(who).is_empty()
```
jordy25519, line 856: would having locks on staking asset prevent dust clearing from other assets with this check?

prml/generic-asset/src/lib.rs
```diff
@@ -185,15 +185,17 @@ pub use types::*;
...
-#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
-pub enum AccountData {
-	None,
-	Exists,
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, Default)]
+pub struct AccountData<AssetId: Ord> {
+	/// Contains asset ids whose balances are greater than existential deposit.
+	significant_assets: BTreeSet<AssetId>,
+	/// True if the account should persist even with dust balances
+	persistent: bool,
```
jordy25519: what is the reason for adding this flag `persistent`?

prml/generic-asset/src/lib.rs
```diff
@@ -185,15 +185,17 @@ pub use types::*;
...
 
 /// Minimal information for an account.
-#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug)]
-pub enum AccountData {
-	None,
-	Exists,
+#[derive(Encode, Decode, Clone, PartialEq, Eq, RuntimeDebug, Default)]
+pub struct AccountData<AssetId: Ord> {
+	/// Contains asset ids whose balances are greater than existential deposit.
+	significant_assets: BTreeSet<AssetId>,
```
jordy25519: would recommend changing `significant` to `active` or `existing_assets` / something related to existence

prml/generic-asset/src/lib.rs
```diff
@@ -185,15 +185,17 @@ pub use types::*;
...
+	/// True if the account should persist even with dust balances
+	persistent: bool,
 }
 
-impl Default for AccountData {
-	fn default() -> Self {
-		AccountData::None
+impl<AssetId: AtLeast32BitUnsigned + Copy> AccountData<AssetId> {
+	fn is_significant(&self) -> bool {
+		!self.significant_assets.is_empty() || self.persistent
```
jordy25519: assuming 'persistent' is required, makes sense to short circuit
```suggestion
	fn exists(&self) -> bool {
		self.persistent || !self.significant_assets.is_empty()
```

prml/generic-asset/src/lib.rs
```diff
@@ -213,7 +215,7 @@ pub trait Config: frame_system::Config {
 	type Event: From<Event<Self>> + Into<<Self as frame_system::Config>::Event>;
 
 	/// The means of storing the balances of an account.
```
jordy25519: should update the doc comment to reflect its real purpose in this module

prml/generic-asset/src/lib.rs
```diff
@@ -725,10 +740,8 @@ impl<T: Config> Module<T> {
 
 		let original_free_balance = Self::free_balance(asset_id, beneficiary);
 		let new_free_balance = original_free_balance + slash;
-		Self::set_free_balance(asset_id, beneficiary, new_free_balance);
-
 		let new_reserve_balance = b - slash;
-		Self::set_reserved_balance(asset_id, who, new_reserve_balance);
+		Self::set_balances(asset_id, who, new_free_balance, new_reserve_balance);
```
jordy25519: is `who` and `beneficiary` guaranteed to always be the same?
before it was:
```rust
		Self::set_free_balance(asset_id, beneficiary, new_free_balance);
		Self::set_reserved_balance(asset_id, who, new_reserve_balance);
````

File: prml/generic-asset/src/lib.rs [hunk 9]
```diff
@@ -802,19 +815,122 @@ impl<T: Config> Module<T> {
...
+					}
+				}
+			};
+			<Result<_, &'static str>>::Ok(maybe_account.clone())
+		})
+		.unwrap_or_default()
+	}
+
+	/// Remove an account's whole trace from the storage
+	fn purge(who: &T::AccountId) {
```
jordy25519: shouldn't purging happen for a particular asset when it's own balance goes to dust? iterating every asset in existence doesn't seem right.
when no assets are left then we could purge the AccountData storage also.

prml/generic-asset/src/lib.rs
aliXsed: It can, but it could complicate the implementation. I have supposed when an account has got any reserved or locked balance, it means it still needs to exist to interact with the modules which did the reserve or lock operation. So it's important that they exist to clear up with the interactors.

prml/generic-asset/src/lib.rs
aliXsed: The naming came from similar Substrate balances functions. I will give it another thought.

prml/generic-asset/src/lib.rs
aliXsed: Dust clearing happens only when we remove an account from the account store. When an account should exist for one asset, it will be able to take care of its own dust.

prml/generic-asset/src/lib.rs
aliXsed: It will be useful for endowed accounts. The accounts that should exist even with dust.

prml/generic-asset/src/lib.rs
aliXsed: `active` would refer to the operations happening in which an account is involved. `existing_assets` would also be confusing as other insignificant assets also exist until there's no asset with significant amounts.

prml/generic-asset/src/lib.rs
aliXsed: Will do that.

prml/generic-asset/src/lib.rs
aliXsed: That would be a different design. I had assumed we should purge an account and its storage only when all the assets are dust. Before that we'd only need to keep the trace of significant assets. This was based on the description in https://github.com/plugblockchain/plug-blockchain/issues/183. The design you are suggesting is less tolerant against dust, I get that. I'm open to this suggestion. Would you want me to change #183 and take this approach?

prml/generic-asset/src/lib.rs
aliXsed: Nice catch. I will fix it right away.

prml/generic-asset/src/lib.rs
jordy25519: Ah I didn't notice this design in #183, It seems better to release storage as we're able to do so

prml/generic-asset/src/lib.rs
aliXsed: Changed it to `existing_assets` to match the new design

prml/generic-asset/src/lib.rs
aliXsed: Done

prml/generic-asset/src/lib.rs
aliXsed: As soon as we address #185 which can be immediately after #183, then we will only check the lock for the corresponding asset.

prml/generic-asset/src/lib.rs
aliXsed: Done

prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +830,114 @@ impl<T: Config> Module<T> {
...
+				}
+			};
+			<Result<_, &'static str>>::Ok(maybe_account.clone())
+		})
+		.unwrap_or_default()
+	}
+
+	/// Remove an account's whole trace from the storage and move the dusts to the treasury
+	fn purge(asset_id: T::AssetId, who: &T::AccountId) {
+		let treasury_account_id: T::AccountId = T::TreasuryModuleId::get().into_account();
```
jordy25519: I think it'd be nicer to pass the imbalance to a trait like `OnPurgedImbalance` or `OnDustImbalance` and let the runtime define how to handle this situation per network

prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +830,114 @@ impl<T: Config> Module<T> {
...
+		let amount = <FreeBalance<T>>::take(asset_id, who);
+		if amount > Zero::zero() {
+			<FreeBalance<T>>::mutate(asset_id, treasury_account_id.clone(), |y| *y += amount);
+		}
+		Self::deposit_event(Event::<T>::Purged(asset_id, who.clone(), amount));
+	}
+
+	/// Call `func` wrapped in the dust check logic and according to `ExistenceRequirement` for the
+	/// `who` account. Update the account store if needed.
+	fn call_with_dust_check(asset_id: T::AssetId, who: &T::AccountId, func: impl FnOnce(), req: ExistenceRequirement) {
```
jordy25519, line 905: 👍

prml/generic-asset/src/lib.rs
```diff
@@ -484,6 +502,8 @@ decl_event! {
 		Minted(AssetId, AccountId, Balance),
 		/// Asset burned (asset_id, account, amount).
 		Burned(AssetId, AccountId, Balance),
+		/// Account has been deleted due to having below existential deposits
```
jordy25519:
```suggestion
		/// Asset balance storage has been reclaimed due to falling below the existential deposit
```

jordy25519 (commented):
quick review, still need to go through tests

prml/generic-asset/src/lib.rs
aliXsed: Done

prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +827,113 @@ impl<T: Config> Module<T> {
...
+	/// true in `asset_is_dust`, if the asset for `who` is at the time of this call considered dust.
+	/// This will purge dust assets for non persistent accounts.
+	fn update_account_store(
+		asset_id: T::AssetId,
+		who: &T::AccountId,
+		asset_is_dust: bool,
+	) -> Option<AccountData<T::AssetId>> {
+		T::AccountStore::try_mutate_exists(who, |maybe_account| {
+			match maybe_account {
+				Some(k) => {
```
jordy25519: nit:
```suggestion
				Some(account) => {
```

prml/generic-asset/src/lib.rs
jordy25519: endowed accounts aren't special, they just get some balance set via genesis config. I'm generally in favour of having less edge cases and treating all accounts the same way for simplicity sake.
if the 'persistent' account were to be cleared it doesn't change anything vs. persistent, it could be recreated by receiving a transfer, it's nonce will still exist in the system module

prml/generic-asset/src/lib.rs
```diff
@@ -484,6 +499,8 @@ decl_event! {
 		Minted(AssetId, AccountId, Balance),
 		/// Asset burned (asset_id, account, amount).
 		Burned(AssetId, AccountId, Balance),
+		/// Asset balance storage has been reclaimed due to falling below the existential deposit
+		Purged(AssetId, AccountId, Balance),
```
jordy25519: something like this will be nicer for dapps/user
```suggestion
		DustReclaimed(AssetId, AccountId, Balance),
```

jordy25519 (commented):
generally lgtm! I don't think the persistent flag has any usefulness

jordy25519 (commented):
looks good.
Need one more test for account transfer an amount < existential deposit should fail

prml/generic-asset/src/lib.rs
```diff
@@ -802,19 +820,110 @@ impl<T: Config> Module<T> {
...
+	fn update_account_store(asset_id: T::AssetId, who: &T::AccountId, asset_is_dust: bool) {
+		let _ = T::AccountStore::try_mutate_exists(who, |maybe_account| {
+			match maybe_account {
+				Some(account) => {
+					if asset_is_dust {
+						account.existing_assets.remove(&asset_id);
+						Self::purge(asset_id, who);
+					} else {
+						account.existing_assets.insert(asset_id);
+					}
```
jordy25519: Looks like the system pallet can help (inspiration from here: https://github.com/paritytech/substrate/blob/3efe87daba09be18b3596e075c3733fe47bafdd5/frame/balances/src/lib.rs#L1074):
```suggestion
						Self::purge(asset_id, who);
						// GA may have the responsibility to keep an account 'alive'
						// if there are no consumers or another provider for the account exists, then we can release the storage safely
						if !account.should_exist() && (!frame_system::Pallet::<T>::is_provider_required(who) || frame_system::Pallet::<T>::providers(who) > One::one()) {
							*maybe_account = None
						}
					} else {
						account.existing_assets.insert(asset_id);
					}
```

aliXsed: The problem was not just with removing the account (`*maybe_account = None`). Any change that would leave the account value to be the same as the default value would fail when a provider is required. That's why we need something like the flag `exists` that would be different than default for the case that we want to remove the last dust asset from existing_assets but leave the account for its consumer. The flag solution is efficient and better than coupling Generic Asset more and more with the system module.

jordy25519: Makes sense. Allowing account to be reclaimed by system (i.e. nonce reaped) seems like too much complexity vs. benefit, have proposed preventing it when GA is the account store in #190. Can tackle it separately if we find the need for it in future

prml/generic-asset/src/tests.rs
```diff
@@ -204,6 +209,313 @@ fn transferring_less_than_one_unit_should_fail() {
...
+
+		// TODO Enable the following check after https://github.com/plugblockchain/plug-blockchain/issues/191
+		// assert!(!System::account_exists(&BOB));
+
+		assert!(!<FreeBalance<Test>>::contains_key(STAKING_ASSET_ID, &BOB));
+	});
+}
+
+#[test]
+fn an_account_with_a_consumer_should_persist_in_system_even_when_ga_not_providing_it() {
```
jordy25519: test name should change since GA will always provide it

jordy25519 (commented):
what happens when a dust amount is transferred to a non-existent account?

prml/generic-asset/src/tests.rs
aliXsed: > what happens when a dust amount is transferred to a non-existent account?

An account with a dust balance will be created. But we don't know if the account is going to always have a dust balance. A couple of dust balance transfer can make an account to have a significant balance. We don't need to stop that to happen. As long as the user has paid the fee for transfer we can be open to the interaction. We uniformly apply the policy to purging dusts to initiators of an interaction or a runtime upgrade.

aliXsed:
> what happens when a dust amount is transferred to a non-existent account?

@holygits An account with a dust balance will be created. But we don't know if the account is going to always have a dust balance. A couple of dust balance transfer can make an account to have a significant balance. We don't need to stop that to happen. As long as the user has paid the fee for transfer we can be open to the interaction. We uniformly apply the policy of purging dusts to initiators of an interaction or a runtime upgrade.

prml/generic-asset/src/tests.rs
aliXsed: Done

jordy25519:
> @holygits An account with a dust balance will be created. But we don't know if the account is going to always have a dust balance. A couple of dust balance transfer can make an account to have a significant balance. We don't need to stop that to happen. As long as the user has paid the fee for transfer we can be open to the interaction. We uniformly apply the policy of purging dusts to initiators of an interaction or a runtime upgrade.

Can you add a test for this please

jordy25519 (approved):
lgtm!

KarishmaBothara (approved):

PR #190, opened 2021-03-31
183 suggestions

jordy25519:
- Always provide accounts, allowing nonce to be reaped is too complex vs. benefits at this point
- Make on runtime_upgrade more storage efficient
- Make make_transfer more efficient and prevent creating dust accounts via transfer
- exists -> tombstone is common term in dbs to reflect this situation (non-existent vs removed)
- Add missing tests for lock transfers

### DIFF
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -815,6 +815,7 @@ impl<T: Config> Module<T> {
 		AssetMeta::<T>::iter().collect()
 	}
 
+	/// Return true if the specified asset of `who` is considered dust (insignificant).
 	fn is_dust(asset_id: T::AssetId, who: &T::AccountId) -> bool {
 		let existential_deposit = AssetMeta::<T>::get(asset_id).existential_deposit();
 		// If for an asset, there is enough deposit above the defined existential deposit, it will not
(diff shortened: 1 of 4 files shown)


prml/generic-asset/src/lib.rs
```diff
@@ -864,6 +882,13 @@ impl<T: Config> Module<T> {
 				Some(account) => {
 					if asset_is_dust {
 						account.existing_assets.remove(&asset_id);
+						if !account.should_exist() {
+							// the account has no more asset balances and is now unused by this module.
+							// mark it as effectively dead with tombstone.
+							// NOTE: GA will continue to act as a provider for the account, preventing it from being reaped by the system module.
+							// see: https://github.com/plugblockchain/plug-blockchain/issues/189
```
jordy25519, line 889: making it explicit

prml/generic-asset/src/lib.rs
```diff
@@ -647,18 +661,22 @@ impl<T: Config> Module<T> {
...
-				|| {
-					<FreeBalance<T>>::insert(asset_id, from, &new_from_balance);
-				},
-				req,
-			);
-			Self::set_free_balance(asset_id, to, new_to_balance);
+			<FreeBalance<T>>::insert(asset_id, from, &new_from_balance);
+			<FreeBalance<T>>::insert(asset_id, to, &new_to_balance);
+			// `new_from_balance` could be dust
+			if new_from_balance < existential_deposit && req == ExistenceRequirement::AllowDeath && Self::reserved_balance(asset_id, from).is_zero() {
```
holygits, line 677: only pay for dust checks in the worst case

PR #192, opened 2021-04-06
Feature/185 ga locks

### LINKED ISSUE #185: Place the locks for a specifed asset

aliXsed:
Closes #185

### DIFF
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -470,7 +470,7 @@ decl_storage! {
 
 		/// Any liquidity locks on some account balances.
 		pub Locks get(fn locks):
-			map hasher(blake2_128_concat) T::AccountId => Vec<BalanceLock<T::Balance>>;
+			double_map hasher(twox_64_concat) T::AssetId, hasher(blake2_128_concat) T::AccountId => Vec<BalanceLock<T::Balance>>;
 
 		/// The identity of the asset which is the one that is designated for the chain's staking system.
 		pub StakingAssetId get(fn staking_asset_id) config(): T::AssetId;
@@ -827,11 +827,11 @@ impl<T: Config> Module<T> {
 			return Ok(());
 		}
 
-		let locks = Self::locks(who);
+		let locks = Self::locks(asset_id, who);
 		if locks.is_empty() {
 			return Ok(());
 		}
-		if Self::locks(who)
+		if locks
 			.into_iter()
 			.all(|l| new_balance >= l.amount || !l.reasons.intersects(reasons))
 		{
@@ -853,7 +853,7 @@ impl<T: Config> Module<T> {
 		// should be kept for the clearance of those operations and thus is not dust.
 		<FreeBalance<T>>::get(asset_id, who) < existential_deposit
 			&& <ReservedBalance<T>>::get(asset_id, who).is_zero()
-			&& Self::locks(who).is_empty()
+			&& Self::locks(asset_id, who).is_empty()
 	}
 
 	/// Update the account of `who` in the account store based on the current asset status. Pass
@@ -891,7 +891,7 @@ impl<T: Config> Module<T> {
 
 	/// Remove an asset for an account and pass a non-zero imbalance to dust imbalance handler.
 	fn purge(asset_id: T::AssetId, who: &T::AccountId) {
-		<Locks<T>>::remove(who);
+		<Locks<T>>::remove(asset_id, who);
 		<ReservedBalance<T>>::remove(asset_id, who);
 		let amount = <FreeBalance<T>>::take(asset_id, who);
 		if amount > Zero::zero() {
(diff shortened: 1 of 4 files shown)


jordy25519 (commented):
changes lgtm.
needs a step for migrating locks on upgrade
since we're here could you add the tests for transfers with locked amounts?

prml/generic-asset/src/lib.rs
```diff
@@ -498,6 +537,14 @@ decl_storage! {
 	}
 }
 
+fn migrate_locks<T: Config, U: IterableStorageMap<T::AccountId, Vec<BalanceLock<T::Balance>>>>(
+	old_locks_iter: U::Iterator,
+) {
+	old_locks_iter.for_each(|(account_id, locks)| {
+		<Locks<T>>::insert(<StakingAssetId<T>>::get(), account_id, locks);
```
jordy25519: does this query storage `<StakingAssetId<T>>::get()`? if so please move outside the loop

prml/generic-asset/src/tests.rs
```diff
@@ -542,6 +560,53 @@ fn on_runtime_upgrade() {
 	});
 }
 
+#[test]
+fn migrate_locks() {
```
jordy25519: would be good to get integration test also, `is_dust()` doesn't check the current state of the locks anymore.

I think it's fine because `free_balance > ED` should always be true for locked balances but `locks.is_empty()` will always be true

prml/generic-asset/src/lib.rs
```diff
@@ -429,11 +429,45 @@ decl_module! {
 				}
 			});
 			dust_accounts.iter().for_each(|(asset_id, account_id)| Self::purge(*asset_id, account_id));
+
+			if StorageVersion::get() == Releases::V0 as u32 {
+				StorageVersion::put(Releases::V1 as u32);
```
jordy25519, line 424: this version check might aswell be over the dust checks also

prml/generic-asset/src/tests.rs
aliXsed: I will do the migration first to make sure is_dust going to work correctly.

prml/generic-asset/src/lib.rs
aliXsed: Done

prml/generic-asset/src/lib.rs
aliXsed: Done

prml/generic-asset/src/lib.rs
```diff
@@ -420,20 +420,45 @@ decl_module! {
...
+
+				let mut dust_accounts = <Vec<(T::AssetId, T::AccountId)>>::new();
+				<FreeBalance<T>>::iter().for_each(|(asset_id, account_id, _)|{
+					let is_dust = Self::is_dust(asset_id, &account_id);
+					Self::update_account_store(asset_id, &account_id, is_dust);
+					if is_dust {
+						dust_accounts.push((asset_id, account_id));
+					}
+				});
+				dust_accounts.iter().for_each(|(asset_id, account_id)| Self::purge(*asset_id, account_id));
```
jordy25519, line 436: could just purge inline instead of allocating a new vec (kind of worried this whole process will be slow on mainnet)

```suggestion
						Self::purge(*asset_id, account_id);
					}
				});
```

prml/generic-asset/src/lib.rs
```diff
@@ -812,26 +863,25 @@ impl<T: Config> Module<T> {
...
-
-		let locks = Self::locks(who);
+		let locks = Self::locks(asset_id, who);
 		if locks.is_empty() {
 			return Ok(());
 		}
-		if Self::locks(who)
+		if locks
 			.into_iter()
 			.all(|l| new_balance >= l.amount || !l.reasons.intersects(reasons))
```
jordy25519, line 886: I don't think now is the time to solve but it seems weird that
2 separate locks of value 10 would allow a `new_balance` of 11 to happen, would expect `new_balance >= sum(locks)`

jordy25519 (approved):
looks great

prml/generic-asset/src/lib.rs
```diff
@@ -498,6 +528,27 @@ decl_storage! {
...
+			trait Store for Module<T: Config> as GenericAsset {
+				pub Locks get(fn locks):
+					map hasher(blake2_128_concat) T::AccountId => Vec<BalanceLock<T::Balance>>;
+			}
+		}
+	}
+
+	let staking_asset_id = <StakingAssetId<T>>::get();
+	let old_locks = <inner::Locks<T>>::drain().collect::<Vec<_>>();
+	old_locks.iter().for_each(|(account_id, locks)| {
```
jordy25519, line 547: this would take elements 1 by 1 vs. allocate a new vec

```suggestion
	<inner::Locks<T>>::drain().iter().for_each(|(account_id, locks)| {
```

KarishmaBothara (approved):

prml/generic-asset/src/lib.rs
aliXsed: I realised Locks<T> and inner::Locks<T> even though they have different layouts and different names, they are internally based on the same storage which is viewed differently. It wasn't sound to insert to Locks while inner::Locks was being drained. It could actually make the integration test fail. So I had to drain the previous content of Locks in a separate step.

prml/generic-asset/src/lib.rs
aliXsed: It is. I create an issue so we revisit this later.

aliXsed: https://github.com/plugblockchain/plug-blockchain/issues/193

prml/generic-asset/src/lib.rs
aliXsed: Purge internally changes FreeBalance and so not safe to be called while iterating on FreeBalance. We will test the speed on Nikau.

PR #197, opened 2021-04-09
Fix minimum_balance of MultiCurrencyAccounting and Currency for GA

aliXsed:
It should return the existential deposit of the relevant asset. One test is added and another test is amended so they maintain this requirement.


KarishmaBothara (approved):
lgtm as non rust expert.

jordy25519 (approved):

PR #204, opened 2021-05-20
Fixing initial issuance decimal place amount

### LINKED ISSUE #178: GA create total supply use decimal places
When creating a GA, initial issuance should be the whole number of tokens to create and the total supply should handle the decimal places i.e

`total_supply = initial_issuance x 10.pow(decimal_places)`

this will be a much simpler API to use client side

while we're here, limit decimal places to `<= 18` seems sensible

JasonTulp:
Closes #178

### DIFF
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -279,7 +279,7 @@ decl_error! {
 		InitialIssuanceTooLarge,
 		/// The multiplication of these two values has lead to the oversaturation of the unsigned integer.
 		MultiplicationOverSaturated,
-		
+
 	}
 }
 
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -668,6 +668,10 @@ impl<T: Config> Module<T> {
 			!info.existential_deposit().is_zero(),
 			Error::<T>::ZeroExistentialDeposit
 		);
+		let decimal_offset = 10u128
+			.checked_pow(info.decimal_places().into())
+			.ok_or(Error::<T>::DecimalTooLarge)?;
+
 		let asset_id = if let Some(asset_id) = asset_id {
 			ensure!(!asset_id.is_zero(), Error::<T>::AssetIdExists);
 			ensure!(!<TotalIssuance<T>>::contains_key(asset_id), Error::<T>::AssetIdExists);
@@ -682,10 +686,6 @@ impl<T: Config> Module<T> {
 
 		let account_id = from_account.unwrap_or_default();
 		let permissions: PermissionVersions<T::AccountId> = options.permissions.clone().into();
-
-		let decimal_offset = 10u128
-			.checked_pow(info.decimal_places().into())
-			.ok_or(Error::<T>::DecimalTooLarge)?;
 
 		let total_issuance: u128 = decimal_offset
 			.checked_mul(
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -273,9 +273,9 @@ decl_error! {
 		ZeroExistentialDeposit,
 		/// There is no such account id in the storage.
 		AccountIdNotExist,
-		/// The integer for Decimal_Places is too large for conversion into u128.
+		/// The integer for decimal places is too large for conversion into u128.
 		DecimalTooLarge,
-		/// The integer for Initial_Issuance is too large for conversion into u128.
+		/// The integer for initial issuance is too large for conversion into u128.
 		InitialIssuanceTooLarge,
 
 	}
@@ -671,7 +671,7 @@ impl<T: Config> Module<T> {
 			.checked_pow(info.decimal_places().into())
 			.ok_or(Error::<T>::DecimalTooLarge)?
 			.unique_saturated_into();
-		// Assuming that Balance is always u128 or larger. Implemented in this way for practicality
+		// Assuming that Balance is u128 or less. Implemented in this way for practicality
 		let total_issuance: T::Balance = decimal_factor
 			.checked_mul(&options.initial_issuance)
 			.ok_or(Error::<T>::InitialIssuanceTooLarge)?;
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -154,7 +154,8 @@
 use codec::{Codec, Decode, Encode, FullCodec};
 
 use sp_runtime::traits::{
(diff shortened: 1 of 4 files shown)


prml/generic-asset/src/mock.rs
```diff
@@ -47,7 +47,7 @@ pub const TEST2_ASSET_ID: u32 = 16004;
 // default next asset id
 pub const ASSET_ID: u32 = 1000;
 // initial issuance for creating new asset
-pub const INITIAL_ISSUANCE: u64 = 1000;
+pub const INITIAL_ISSUANCE: u64 = 1000000;
```
jordy25519:
```suggestion
pub const INITIAL_ISSUANCE: u64 = 1_000_000;
```

prml/generic-asset/src/lib.rs
```diff
@@ -271,6 +273,11 @@ decl_error! {
 		ZeroExistentialDeposit,
 		/// There is no such account id in the storage.
 		AccountIdNotExist,
+		/// The integer for Decimal_Places is too large for conversion into u128.
+		DecimalTooLarge,
+		/// The integer for Initial_Issuance is too large for conversion into u128.
```
jordy25519:
```suggestion
		/// The integer for initial issuance is too large for conversion into u128.
```

prml/generic-asset/src/lib.rs
```diff
@@ -271,6 +273,11 @@ decl_error! {
 		ZeroExistentialDeposit,
 		/// There is no such account id in the storage.
 		AccountIdNotExist,
+		/// The integer for Decimal_Places is too large for conversion into u128.
```
jordy25519:
```suggestion
		/// The integer for decimal places is too large for conversion into u128.
```

prml/generic-asset/src/lib.rs
```diff
@@ -659,6 +666,16 @@ impl<T: Config> Module<T> {
 			!info.existential_deposit().is_zero(),
 			Error::<T>::ZeroExistentialDeposit
 		);
+
+		let decimal_factor: T::Balance = 10u128
+			.checked_pow(info.decimal_places().into())
+			.ok_or(Error::<T>::DecimalTooLarge)?
+			.unique_saturated_into();
+		// Assuming that Balance is always u128 or larger. Implemented in this way for practicality
```
jordy25519:
```suggestion
		// Assuming that Balance is u128 or less. Implemented in this way for practicality
```

prml/generic-asset/src/tests.rs
```diff
@@ -34,9 +34,22 @@ use frame_support::{
...
+fn asset_options(permissions: PermissionLatest<u64>, decimal_place: u8) -> AssetOptions<u64, u64> {
+	let decimal_factor = 10u128.saturating_pow(decimal_place.into());
 	AssetOptions {
-		initial_issuance: INITIAL_ISSUANCE,
+		initial_issuance: (INITIAL_ISSUANCE as u128 / decimal_factor) as u64,
+		permissions,
+	}
+}
+
+fn asset_options_set_initial_issuance(
```
jordy25519: I don't think this function is helpful for the tests, better to be explicit about the real total issuance value in the test, see "explicit is better than implicit"

jordy25519 (commented):
generally, lgtm minus few nits

prml/generic-asset/src/tests.rs
```diff
@@ -1546,18 +1576,101 @@ fn create_asset_should_create_a_user_asset() {
...
+			web3_asset_info.clone()
+		));
+
+		assert_eq!(<AssetMeta<Test>>::get(ASSET_ID), web3_asset_info);
+		assert_eq!(GenericAsset::total_issuance(&ASSET_ID), INITIAL_ISSUANCE);
+	});
+}
+
+#[test]
+fn create_asset_should_add_decimal_places_maximum() {
```
jordy25519: what behaviour does this test ensure? / what does 'maximum' mean here?

jordy25519 (approved):
one last thing but other than that lgtm

prml/generic-asset/src/tests.rs
JasonTulp: The purpose of this test was to ensure that it is possible to create an asset with 18 decimal places, (Which is the theoretical maximum we will need) Maybe not needed but thought I would test both the upper and lower bounds for successful asset creation.

JasonTulp: I could change the name of the Test to something like the following: create_asset_should_work_with_max_decimal_places

jordy25519: gotcha, I see the `initial_issuance` value here as a kind of [magic number](https://en.wikipedia.org/wiki/Magic_number_(programming)) `10000000000000000000` not any kind of real maximum, and why set it so large then divide it by `decimal_factor`? just use the constant value you'd like to test

JasonTulp: I understand that this test could be simplified by setting `initial_issuance = 10` and skipping the calculation and I am happy to make this change, however for the sake of readability it may get confusing why this value then matches `10*10^18` in the `assert_eq!` test on line 1656. By showing the calculations it makes it a little clearer for anyone down the line as to what is actually happening. 
I would have preferred to use the constant `INITIAL_ISSUANCE = 1_000_000,` however for this test it would need to be at least 18 0's long to allow for the maximum of `decimal_places = 18`, and by showing the calculation of `initial_issuance / decimal_factor` the reasoning for the massive number is made clearer.
That being said, I can see that it is two extra steps and I am happy to add hardcoded values in both the `AssetOptions` and final test.

prml/generic-asset/src/tests.rs
```diff
@@ -1546,18 +1576,101 @@ fn create_asset_should_create_a_user_asset() {
...
+			ALICE,
+			AssetOptions {
+				initial_issuance: (initial_issuance / decimal_factor as u64),
+				permissions,
+			},
+			web3_asset_info.clone(),
+		));
+
+		assert_eq!(<AssetMeta<Test>>::get(ASSET_ID), web3_asset_info);
+		assert_eq!(GenericAsset::total_issuance(&ASSET_ID), initial_issuance);
```
jordy25519, line 1657: so here's another way to write this test which is abit clearer on intention
```suggestion
		let initial_issuance = 1_u64;
		let decimal_factor = 10_u128.saturating_pow(web3_asset_info.decimal_places().into());
		let total_issuance = initial_issuance.saturating_mul(decimal_factor);

		assert_ok!(GenericAsset::create(
			Origin::root(),
			ALICE,
			AssetOptions {
				initial_issuance,
				permissions,
			},
			web3_asset_info.clone(),
		));

		assert_eq!(<AssetMeta<Test>>::get(ASSET_ID), web3_asset_info);
		assert_eq!(GenericAsset::total_issuance(&ASSET_ID), total_issuance);
```

PR #205, opened 2021-05-23
For a plug node, replace pallet_balances with generic_asset.

aliXsed:
Replace pallet Balances with Generic Asset for a plug node and do another round of benchmarking generic asset. The command used to benchmark is:
`cargo run --release --features runtime-benchmarks -p node-cli -- benchmark --chain dev --steps 50 --repeat 2 --pallet prml_generic_asset --extrinsic "*" --raw --execution=wasm --wasm-execution=compiled --output generic_asset.rs`

### DIFF
--- a/bin/node/executor/tests/submit_transaction.rs
+++ b/bin/node/executor/tests/submit_transaction.rs
@@ -17,7 +17,7 @@
 
 use std::sync::Arc;
 use node_runtime::{
-	Executive, Indices, Runtime, UncheckedExtrinsic,
+	Executive, GenericAsset, Indices, Runtime, UncheckedExtrinsic,
 };
 use sp_application_crypto::AppKey;
 use sp_core::{
@@ -92,7 +92,11 @@ fn should_submit_signed_transaction() {
 	t.execute_with(|| {
 		let results = Signer::<Runtime, TestAuthorityId>::all_accounts()
 			.send_signed_transaction(|_| {
-				pallet_balances::Call::transfer(Default::default(), Default::default())
+				prml_generic_asset::Call::transfer(
+					GenericAsset::spending_asset_id(),
+					Default::default(),
+					Default::default(),
+				)
 			});
 
 		let len = results.len();
@@ -124,7 +128,11 @@ fn should_submit_signed_twice_from_the_same_account() {
 	t.execute_with(|| {
 		let result = Signer::<Runtime, TestAuthorityId>::any_account()
 			.send_signed_transaction(|_| {
-				pallet_balances::Call::transfer(Default::default(), Default::default())
+				prml_generic_asset::Call::transfer(
+					GenericAsset::spending_asset_id(),
+					Default::default(),
+					Default::default(),
+				)
 			});
 
 		assert!(result.is_some());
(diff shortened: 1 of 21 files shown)


bin/node/runtime/src/lib.rs
```diff
@@ -1110,6 +1091,7 @@ construct_runtime!(
 		Lottery: pallet_lottery::{Module, Call, Storage, Event<T>},
 		Gilt: pallet_gilt::{Module, Call, Storage, Event<T>, Config},
 		Attestation: prml_attestation::{Module, Call, Storage, Event<T>},
+		GenericAsset: prml_generic_asset::{Module, Call, Storage, Event<T>, Config<T>},
```
jordy25519: @aliXsed try putting this before Staking in the order

aliXsed: Done

bin/node/executor/tests/submit_transaction.rs
```diff
@@ -251,7 +275,8 @@ fn submitted_transaction_should_be_valid() {
 		// add balance to the account
 		let author = extrinsic.signature.clone().unwrap().0;
 		let address = Indices::lookup(author).unwrap();
-		let data = pallet_balances::AccountData { free: 5_000_000_000_000, ..Default::default() };
+		// TODO the account should have 5_000_000_000_000 of spending asset free
```
jordy25519, line 278: did you mean to come back to this?

prml/generic-asset/src/lib.rs
```diff
@@ -318,9 +318,27 @@ decl_module! {
...
+		/// 99% of the time you want [`transfer`] instead.
+		///
+		/// [`transfer`]: struct.Pallet.html#method.transfer
+		/// # <weight>
+		/// - Cheaper than transfer because account cannot be killed.
+		/// - Base Weight: 51.4 µs
+		/// - DB Weight: 1 Read and 1 Write to dest (sender is in overlay already)
+		/// #</weight>
+		#[weight = T::WeightInfo::transfer_keep_alive()]
+		pub fn transfer_keep_alive(origin, #[compact] asset_id: T::AssetId, to: T::AccountId, #[compact] amount: T::Balance)
```
jordy25519, line 336: interesting, wouldn't everyone use this option instead of `transfer` since it is cheaper? seems to undermine the existential deposit stuff

jordy25519 (approved):
lgtm in general. ` transfer_keep_alive` raises some questions about why we implemented ED rules at all

bin/node/executor/tests/submit_transaction.rs
aliXsed: I think as the tests are passing I can now remove this TODO note.

prml/generic-asset/src/lib.rs
aliXsed: Yeah, but that's coming from substrate and they have used it in their test.

bin/node/executor/tests/submit_transaction.rs
aliXsed: I will remove it in my next PR.

KarishmaBothara (approved):

prml/generic-asset/src/lib.rs
jordy25519: I think we should mark this `#[test]` or remove ED rules

PR #206, opened 2021-05-24
Feature/202 relax create asset restrictions

### LINKED ISSUE #202: Relax create asset restrictions
Allow 'create asset' calls from any account.
Created reserved can stay sudo or require deposit

'create_reserved' issues asset IDs in ranges 0-1000 (future-proofing for assets which might have system privilege)
'create' issues asset IDs in range > 1000, anyone can create

---
At the moment creating a new generic asset requires sudo i.e governance to perform.

It's possible a 'generic assets' council exists to vote on acceptance of new tokens based on some simple application process
This is secure but maybe abit too bureaucratic.

Other options include requiring a one-time CENNZ deposit to create to prevent easily making spammy assets

JasonTulp:
Closes #202

### DIFF
--- a/prml/generic-asset/src/lib.rs
+++ b/prml/generic-asset/src/lib.rs
@@ -302,7 +302,9 @@ decl_module! {
 			options: AssetOptions<T::Balance, T::AccountId>,
 			info: AssetInfo<T::Balance>,
 		) -> DispatchResult {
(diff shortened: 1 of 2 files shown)


jordy25519 (approved):
lgtm!

Issue #210, opened 2021-05-26
transfer keep alive


jordy25519:
'Transfer keep alive' undermines ED rules

https://github.com/plugblockchain/plug-blockchain/pull/205#discussion_r640126855

Issue #191, opened 2021-04-05
Remove accounts purged in GA from the account store


aliXsed:
With the current implementation of providing accounts in generic asset, even when an account is removed from the generic asset storage, generic asset keeps to be providing that account from the frame system's point of view. That's because reaping an account and thus its nonce from the account store would incur some complexity for the case the account is again created. As this may be a potential storage bloating to keep the dead accounts in the frame system, we should consider to remove them in a safe way later.
Trace the todo comments in the code with this issue number 191 as an initial step.

PR #213, opened 2021-06-13
Simplify balance ED rules

### LINKED ISSUE #188: Investigate reentrance safety of Generic Asset functions
Similar to frame/balances/src/tests_reentrancy.rs in https://github.com/paritytech/substrate/pull/8087/files, Add tests to GA that would prove the reentrance safety of the functions there.

jordy25519:
Changes:
1) Release balance related storage from state as soon as possible (free, reserve, locks).
2) GA will act as a perpetual 'provider' of an account when it receives some free balance i.e. it will never `dec_provider`.

This is because the relationship and optimisations between system + balances are different from system + GA.  
We don't need require `AccountData` reaping at this stage

closes #188 #191 #210

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,7 +60,7 @@ commands:
   cargo-run-test:
     steps:
       - run:
-          command:
+          command: |
             cargo --version
             cargo test --release --all
           no_output_timeout: 30m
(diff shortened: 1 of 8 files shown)


JasonTulp (approved):
lgtm!

jordy25519:
closes #188 #191 #210

